PR #31, opened 2015-07-10
Add vagrant support

imobach:
A Vagrantfile is provided to set up a development environment. It uses a 'quick and dirty' script that will be converted in something cleaner in a near future.

Just rename `Vagrantfile.sample` to `Vagrantfile` and type `vagrant up`. It could take a while so, please, be patient.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
-node_modules/
-bower_components/
+node_modules
+bower_components
 .sass-cache/
 .tmp/
 dist/
 src/app/config.local.json
 src/app/config.js
+Vagrantfile
+.vagrant/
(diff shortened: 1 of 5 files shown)


ancorgs:
It almost worked for me. I had some problems because `/home/vagrant/jangouts` was owned by root. After a `chown -R` everything worked.

ancorgs:
Using libvirt, btw (just in case it matters)

imobach:
As it looks like a problem with NFS configuration (and I can't debug it right now because of NFS is not working for me at this time), I'm merging this change (works quite ok using VirtualBox).

Thanks for testing!

PR #41, opened 2015-08-18
Add devel_C_C++ pattern when provisioning

imobach:


ancorgs:
Just out of curiosity. Why is it needed?

imobach:
To make sure that we have the tools to compile Node. I've been experimenting with a rather minimal openSUSE box lately and I found that Node can't be compiled.

PR #45, opened 2015-08-20
WIP: New angular-gridster based UI

ancorgs:
Almost there. Things to improve:

 * Everything about the styling and flow of the thumbnails
 * A lot of css (buttons, header, footer, etc).

Anyway, it already looks and works better than before.


imobach:
LGTM

PR #47, opened 2015-08-30
Upstream/toastr

ChrisBr:
Very simple implementation of the timeout. It only notifies you if the last notification happened more than one minute ago.


ChrisBr:
@ancorgs @imobach please review!

src/app/components/feed/feeds.factory.js
```diff
@@ -29,6 +29,7 @@
       this.isPublisher = attrs.isPublisher || false;
       this.isLocalScreen = attrs.isLocalScreen || false;
       this.isIgnored = attrs.ignored || false;
+      this.notifierTimeout = 0;
```
ancorgs: I wouldn't discard more notifications in the future, so I'd use a less generic name. Let's say something like `speakMutedTimeout`

ancorgs:
I miss resetting the status of the timeout if the user unmutes and mutes again.

src/app/components/feed/feeds.factory.js
ChrisBr: Okay!

ChrisBr:
@ancorgs Good point!

ChrisBr:
@ancorgs probably I won't have time to finish it before the weekend .. jfyi

ancorgs:
Don't worry. I'm on vacation anyways

ancorgs:
Superseded by #53

PR #53, opened 2015-09-11
Notification for muted speakers

ancorgs:
Based on the work by @ChrisBr on #47. I was thinking that displaying notifications should not be in the feeds factory, but in the jhFeed component. So I moved it... and ended up implementing the missing pieces. So this supersedes #47 and should be ready to merge. Please review.

In the future, it should be doable to add a button to unmute inside the notification, but due to (intentional) restrictions on angular-toastr (see https://github.com/Foxandxss/angular-toastr/issues/3), it's not completely straightforward.


src/app/components/feed/feeds.factory.js
```diff
@@ -109,9 +109,12 @@
        *
        * @param {string} type - "audio" or "video"
        * @param {boolean} enabled
+       * @param {object} options - use the 'after' key to specify a callback
+       *        that will be called after configuring the connection.
        */
-      this.setEnabledChannel = function(type, enabled) {
+      this.setEnabledChannel = function(type, enabled, options) {
         var that = this;
+        if (!options) { options = {}; }
```
imobach: I'd prefer something like:

```javascript
options = options || {};
```

imobach: Yes a minor comment. LGTM.

PR #63, opened 2015-09-24
 Add server-side logging capabilities

imobach:
This is just an experimental logging system. Using RemoteLoggingService you can register client events on the server.

Those events are registered by a simple logger (implemented with [hapi.js](http://hapi.js/)).

This feature is only intended as a debugging tool.


imobach:
This code is available yet in [remote-logging](https://github.com/jangouts/jangouts/tree/remote-logging) branch in upstream.

Issue #52, opened 2015-09-10
Automatically resized feeds in list


ancorgs:
The thumbnails are currently fixed at 100px wide. The plan has always being to have them automatically adjusted between 126 and 80 pixels, based on the available space. With the current flex box, that should be very easy to implement using only css with the `min-width` and `max-width` properties. The problem is that those properties, that work as expected with figures containing images, doesn't seem to work with videos.

Two options:
 * CSS based solution. Not sure if it's possible but maybe we can force the videos to act as images and allow the resizing.
 * JS solution, use this method https://github.com/jangouts/jangouts/blob/master/src/app/components/videochat/jh-video-chat.directive.js#L146 to resize the thumbnails to one of these predefined widths: 80, 96, 112, 126.

ancorgs:
After some css with flexbox, I'd say a CSS solution is not possible. Not because the use of video tags, it's even impossible with images. Flex boxes only take into account the main axis when deciding the size of the elements, and we need to take both into account.

PR #67, opened 2015-09-29
Adapt thumbnails to available space

ancorgs:
Implements #52 

Width goes from 64px to 128px (it was fixed at 100px before)


imobach:
LGTM

PR #81, opened 2015-12-07
Readme update

jreidinger:


jreidinger:
please do not merge yet, as my jangouts instance crash like being on demolition derby

jreidinger:
Can be merged now. Last commit added initial troubleshooting, that I need to find that janus need different certificate names.

jreidinger:
The last strange part is that `adapter.js` is not distributed as firefox debug console print me, but tested that jangouts itself works ( thanks @mvidner for help with testing ).

README.md
```diff
@@ -44,7 +44,7 @@ rooms at ```/etc/janus/janus.plugin.videoroom.cfg``` if desired) with the
 command:
 
 ```sh
-systemctl start janus.service
+sudo systemctl start janus.service
```
jreidinger: maybe we should consider also enable? as after start I forgot to start it again.

README.md
```diff
@@ -123,6 +129,17 @@ instance to use the screen sharing functionality you will have to
 provide HTTPS access to both the files and the Janus gateway, like shown
 in the [deployment instructions](DEPLOYMENT.md).
 
+## Troubleshooting
+
+When jangouts server do not work, there is wide variety of ways how to ways how to debug it.
```
imobach: Looks confusing to me. Moreover, there's no 'jangouts' server (just a Janus server).

README.md
imobach: Yes, I agree.

imobach:
Thanks for checking the README! There's only one minor thing (in the Troubleshooting section). Otherwise, LGTM.

README.md
jreidinger: OK, I will add it

README.md
jreidinger: well, for me jangouts server is whole service including all js and web services. Attaching to janus server, etc. So thats why I use jangouts...just first trouble is janus specific, as it doesn't work.

imobach: I see your point, but it could confuse people about Jangouts having some kind of backend (apart from Janus). Anyway, it's fine with me as it's better than before :)

PR #85, opened 2015-12-08
Use ogg instead of mp3

### LINKED ISSUE #83: Bell sound in proprietary mp3
Some browser without free codecs do not play bell sound as it is in proprietary codec. It should be converted into ogg file.

ancorgs:
The bell is still equally annoying, but this fix #83 and makes the bell ring nicely in my chromium (it didn't work before).


imobach:
LGTM.

jreidinger:
LGTM

PR #89, opened 2015-12-08
Add angular-sanitize version to bower.json

imobach:
Small fix to avoid bower complaining about angular-sanitize version.


jreidinger:
LGTM

imobach:
Thanks!

PR #90, opened 2015-12-08
Disable video

jreidinger:
After testing with @lslezak new behavior is:

1) own video is black one
2) when I speak I do not see black screen but instead other person with video
3) on lada pc he see my picture if I speak ( probably reason is that video is not disabled, it is just "frozen" )

### DIFF
--- a/src/app/components/feed/feeds.factory.js
+++ b/src/app/components/feed/feeds.factory.js
@@ -198,6 +198,12 @@
               });
             }
           });
+          if (type === "video") {
+            var track = stream.getVideoTracks();
+            if (track !== null && track !== undefined){
+              track[0].enabled = enabled;
+            }
+          }
         } else if (type === "audio" && enabled === false) {
           DataChannelService.sendMuteRequest(this);
         }
--- a/src/app/components/feed/feeds.factory.js
+++ b/src/app/components/feed/feeds.factory.js
@@ -199,9 +199,9 @@
             }
           });
           if (type === "video") {
-            var track = stream.getVideoTracks();
-            if (track !== null && track !== undefined){
-              track[0].enabled = enabled;
+            var tracks = stream.getVideoTracks();
+            if (tracks !== null && tracks !== undefined){
+              tracks[0].enabled = enabled;
             }
           }
         } else if (type === "audio" && enabled === false) {
--- a/src/app/components/videochat/jh-video-chat.directive.js
+++ b/src/app/components/videochat/jh-video-chat.directive.js
@@ -122,7 +122,11 @@
           }
           // If current one is still speaking, there is no need to change
           if ( !(current && current.getSpeaking()) ) {
-            vm.highlight.current = FeedsService.speakingFeed() || current || FeedsService.findMain();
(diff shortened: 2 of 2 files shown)


src/app/components/feed/feeds.factory.js
```diff
@@ -198,6 +198,12 @@
               });
             }
           });
+          if (type === "video") {
+            var track = stream.getVideoTracks();
```
ancorgs: Nitpicking: the variable should be called `tracks` (plural)

src/app/components/videochat/jh-video-chat.directive.js
```diff
@@ -122,7 +122,11 @@
           }
           // If current one is still speaking, there is no need to change
           if ( !(current && current.getSpeaking()) ) {
-            vm.highlight.current = FeedsService.speakingFeed() || current || FeedsService.findMain();
+            if (FeedsService.speakingFeed() && FeedsService.speakingFeed().getVideoSubscription()){
```
ancorgs: You should not use `getVideoSubscription` here. That function only works for publisher feeds (i.e. yourself and your screen sharings). Use something like `someFeed.isEnabled("video")` instead. That function will always give you the right value, no matter if it's a publisher or subscriber feed.

ancorgs: Although it's almost impossible, there is a chance than the two consecutive calls to `FeedService.speakingFeed()` return different feeds. Better use a variable, like
```
var speaking = FeedService.speakingFeed();
if (speaking && speaking.isEnabled("video")) {
```

src/app/components/videochat/jh-video-chat.directive.js
```diff
@@ -122,7 +122,11 @@
           }
           // If current one is still speaking, there is no need to change
           if ( !(current && current.getSpeaking()) ) {
```
ancorgs: I wonder, shouldn't we change if `current` disables his/her video while speaking? I think we should. So I would change this to
```javascript
// If current one is still speaking and has not disabled the video, there is no need to change
if ( !(current && current.getSpeaking() && current.isEnabled("video") )) {
```
What do you think?

jreidinger: yes, make sense

jreidinger: question is what happen as current is also Speaking, so it ends up with findMain, which in the end is publisher, as current cannot disable video ( from testing his video is still open and only last picture is frozen )

src/app/components/videochat/jh-video-chat.directive.js
jreidinger: ok

src/app/components/videochat/jh-video-chat.directive.js
ancorgs: Of course if everybody is disabling the video you will end up with some weird situation. We probably need to have placeholders also in the main feed view to handle that. But right now I would consider that a non relevant corner case.

ancorgs:
lgtm

PR #94, opened 2015-12-09
Fix loading templates for angular-extended-notification

### LINKED ISSUE #93: Notifications templates are not loaded when using dist/
The code generated by `gulp build` won't load the [angular-extended-notification templates](https://bitbucket.org/lsystems/angular-extended-notifications).

imobach:
This should fix #93. Copying the templates is surely not the best way but, as angular-extended-notification seems to be not actively developed anymore, it just enough for us.

### DIFF
--- a/src/app/templates/gmail-notification.html
+++ b/src/app/templates/gmail-notification.html
@@ -1,0 +1,57 @@
+<div class="gmail-template">
+  <style scoped>
+    .gmail-template {
+      position: relative;
+      right: 0;
+      left: 0;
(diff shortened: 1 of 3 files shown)


ancorgs:
lgtm

src/app/templates/gmail-notification.html
```diff
@@ -0,0 +1,57 @@
+<div class="gmail-template">
```
jreidinger: actually we need only one, as we do not use gmail style notifications

PR #97, opened 2015-12-09
Fix retrieving user layout settings

ChrisBr:
Changing variable names without testing is always a bad idea ...
Introduced by my previous PR, sorry!


imobach:
No problem!

PR #99, opened 2015-12-10
Improve layout persistance

ChrisBr:
We save it now when quitting the window resize mode and furthermore we show a notification
to the user.


imobach:
LGTM. Please, merge!

Issue #77, opened 2015-11-23
Automatically resized thumbnails are sometimes too big


ancorgs:
The thumbnails of the participants are automatically resized to make sure they all fit in the roster area. But there is a bug somewhere and in some situations the resized thumbnails don't fit and a scroll bar appears (instead of using smaller thumbnails).

ancorgs:
@mdeniz I have been reproducing this. It's very easy to reproduce for me with Firefox. I found that the actual problem is not the thumbnails being too big. It's the browser drawing a scrollbar when it's actually not needed.

I'm not sure if there is something we can do, other than being more conservative with the sizes of the thumbnails to prevent Firefox from miscalculating the scrollbar.

ancorgs:
Hmmm, even being more conservative, Firefox does weird things with the scroll bar. I'd say it's time to take a look into some javascript library to fine-control the scrollbars, line nanoScroller or Tiny Scrollbar, so we have better control on it.

PR #100, opened 2015-12-10
Avoid scrollbars to appear if they aren't needed

mdeniz:
It should fix issue #77. Please review.


src/app/components/videochat/jh-video-chat.directive.js
```diff
@@ -222,18 +222,21 @@
         var w = div.innerWidth();
         var h = div.innerHeight();
         var feedW = thumbnailWidth($(".thumb", div).size(), h, w);
+        if (feedW == 64) {
```
jreidinger: I think this should be documented in code that it is just workaround for broken firefox implementation of auto, which basically do auto hidding

mdeniz:
This PR was reimplemented at Ancor's desk, he will create another to fix the issue. Thanks @jreidinger for the tip.

PR #101, opened 2015-12-10
Make UI responsive again (fix some issues in the header)

### LINKED ISSUE #19: Interface for mobile devices
The current UI is not much friendly for devices with a small screen. Moreover, the CPU of mobile devices do not cope very well with many videos being rendered at the same time. Thus, the mobile interface should reduce the number of live videos, maybe rendering only the active one.

mdeniz:
Now gridster is responsive again and I fixed the arrangement of the buttons in the header to be also responsive. It fixes #19.


imobach:
LGTM. Please, merge!

PR #102, opened 2015-12-10
Replace "Video thumbnails" checkbox with a button only in the topbar (not at the signing form)

mdeniz:
This is how it was before:
video_thumbnails_with_text

Now with the button unchecked:
video_thumbnails_with_button_unchecked

With the button checked:
video_thumbnails_with_button_checked

And the tooltip changed:
video_thumbnails_with_button_tooltip

### DIFF
--- a/src/app/components/user/jh-signin-form.html
+++ b/src/app/components/user/jh-signin-form.html
@@ -27,7 +27,7 @@
         <option value="" disabled selected>Select a room</option>
       </select>
     </div>
-    <jh-thumbnails-mode-button></jh-thumbnails-mode-button>
+    <jh-thumbnails-mode-button text-template="true"></jh-thumbnails-mode-button>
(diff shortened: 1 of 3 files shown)


imobach:
LGTM. Merge, please!

jreidinger:
Is it somehow consistent with checkbox on login screen? Now it is different, which I found a bit inconsistent.

imobach:
@jreidinger Inconsistent in which way?

jreidinger:
that here is button and on login is checkbox. So you need to recognize two things for same functionality

imobach:
I see your point but for me it's also inconsistent to mix icons and texts in the 'menu bar'. And using only texts (doing things the other way around) would waste a lot of space.

So I'm fine with this change as it is (just my personal taste).

PR #103, opened 2015-12-10
Fix resizing of feeds in Firefox

### LINKED ISSUE #77: Automatically resized thumbnails are sometimes too big

ancorgs:
Fix #77


mdeniz:
LGTM :+1:

PR #106, opened 2015-12-11
Doc

jreidinger:
improve a bit documentation now `jsdoc -r src` generate something useful.

### DIFF
--- a/src/app/components/feed/feeds.factory.js
+++ b/src/app/components/feed/feeds.factory.js
@@ -20,20 +20,25 @@
   /**
    * Factory representing a janus feed
    * @constructor
-   * @memberof module:janusHnagouts
+   * @memberof module:janusHangouts
    */
   function feedFactory($timeout, DataChannelService, SpeakObserver) {
     return function(attrs) {
       attrs = attrs || {};
       var that = this;
 
+      /** @var {integer} id of feed */
       this.id = attrs.id || 0;
+      /** @var {string} name of user streaming feed */
       this.display = attrs.display || null;
-      this.connection = attrs.connection || null;
+      /** @var {boolean} flag if feed is publishing one, so one that is send from this pc to others */
       this.isPublisher = attrs.isPublisher || false;
+      /** @var {boolean} flag if feed is local screen sharing feed */
       this.isLocalScreen = attrs.isLocalScreen || false;
+      /** @var {boolean} flag if feed is ignored */
       this.isIgnored = attrs.ignored || false;
 
+      var connection = attrs.connection || null;
       var picture = null;
       var speaking = false;
       var silentSince = Date.now();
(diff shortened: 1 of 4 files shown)


src/app/components/feed/feeds.factory.js
```diff
@@ -16,19 +20,25 @@
...
       this.display = attrs.display || null;
-      this.connection = attrs.connection || null;
+      /** @var {boolean} flag if feed is publishing one, so one that is send from this pc to others */
       this.isPublisher = attrs.isPublisher || false;
+      /** @var {boolean} flag if feed is local screen sharing feed */
       this.isLocalScreen = attrs.isLocalScreen || false;
+      /** @var {boolean} flag if feed is ignored */
       this.isIgnored = attrs.ignored || false;
 
+      var connection = attrs.connection || null;
```
ancorgs: Mmm, you changed `this.connection` into `var.connection`. Are you sure everything still works?

src/app/components/feed/feeds.service.js
```diff
@@ -13,9 +13,14 @@
...
+  /**
+   * Service containing all feeds
+   * @constructor
+   * @memberof module:janusHangouts
+   */
   function FeedsService($q, $window) {
-    this.mainFeed = null;
-    this.feeds = {};
+    var mainFeed = null;
+    var feeds = {};
```
ancorgs: If you change this (which I think is correct), shouldn't you change all the references in the file?

src/app/components/feed/feeds.service.js
```diff
@@ -85,22 +106,34 @@
...
+     * @returns {Array<Feed>} all registered feeds sharing local screen
+     */
     function localScreenFeeds() {
       return _.filter(this.allFeeds(), function (f) {
         return f.isLocalScreen;
       });
     }
 
+    /**
+     * @returns {Array<Feed>} all registered feeds that speaks
```
ancorgs: That's wrong. This returns just one feed. https://lodash.com/docs#find

src/app/components/room/room.service.js
```diff
@@ -15,6 +15,11 @@
       'FeedConnection', 'DataChannelService', 'ActionService', 'jhConfig',
       'ScreenShareService'];
 
+  /**
+   * Service containing all rooms
```
ancorgs: Well, that's not how I would describe this service. It has a function to get the list of rooms, but it is not a collection of items like, for example `FeedsService`.

I would rather say something like "Service to manage the communication with the Janus server".

src/app/components/feed/feeds.factory.js
jreidinger: I grepping lines and do not find any outside usage.

src/app/components/room/room.service.js
jreidinger: If it is service to manage communication with janus, then why it is called Room Service?

src/app/components/feed/feeds.factory.js
jreidinger: ah, doing it next time it is used for main feed once, probably appear after merge. so revert

src/app/components/feed/feeds.service.js
jreidinger: done in next commit

src/app/components/feed/feeds.service.js
jreidinger: fixed

src/app/components/room/room.service.js
ancorgs: It provides functions to enter a room, to publish a new feed in the room, to get the list of participants in a room and attach to their feeds... Everything is related to THE room, which is kind of a singleton entity, since you cannot operate in two rooms at the same time.

At some point (it was not there initially) we added a function to get the list of rooms (in plural) and even a `RoomsFactory` to represent the attributes of rooms, but the name `RoomService` was still good enough to stay (and rather funny). :wink: It's a service to perform operations on the room.

A service created to contain all the rooms (if such thing would exist) would be called `RoomsService` (note the plural).

ancorgs: So maybe "Service to manage the communication with the Janus room".

ancorgs:
LGTM

PR #108, opened 2015-12-11
More robust layout management

ancorgs:


imobach:
LGTM

src/app/components/notifier/notifier.service.js
```diff
@@ -18,7 +18,7 @@
 
     function info(text) {
       notifications.info("Information", text, {
-        duration: 20000,
+        duration: 5000,
```
jreidinger: Reason why I use so big number is to allow user to move mouse and click button, for older users like me it is hard to notice notification, catch mouse and then move it there :)

ancorgs: Yes. That's ok for `MuteNotifier`. This one is regular `Notifier`, where user interaction is not expected.

jreidinger: ah, ok, I miss it, that it is not Mute one.

PR #112, opened 2015-12-21
JSON config

imobach:
Configuration is done through `src/config.json` file. An example is in `src/config.json.sample`. The version will be automatically injected.


ancorgs:
LGTM.

I even tested it. It also copies `config.json.sample` to `dist`, which probably was not intended but is harmless anyways (in fact, we have a complete `app` tree that is probably useless there, so one more file...) :-)

imobach:
@ancorgs Oh, I added the sample file at the end. I'll fix it before merging.

imobach:
Fixed and merged.

PR #113, opened 2015-12-22
Avoid create empty directories with gulp build

imobach:


jreidinger:
LGTM, I am just curious why on such simple operation another lib is needed.

ancorgs:
I was going to comment exactly the same that @jreidinger. It LGTM and I'm surprised how tricky it resulted to be.

imobach:
@ancorgs @jreidinger Maybe there's a simple way :)

PR #125, opened 2016-03-02
Start muted if there are more than 3 participants. This closes #12.

### LINKED ISSUE #12: Start muted if there are more than 3 participants
Like it happens in Google Hangouts when you join a crowded session.

sandr3j:
Audio is now off by default and will be toggled if the number of obtained participants is less than 4 when the user joins.

### DIFF
--- a/src/app/components/feed/connection-config.factory.js
+++ b/src/app/components/feed/connection-config.factory.js
@@ -24,7 +24,7 @@
     return function(pluginHandle, wantedInit, jsep, ok) {
       var current = {};
       var requested = null;
-      var wanted = {audio: true, video: true};
+      var wanted = {audio: false, video: true};
       var okCallback = null;
       _.assign(wanted, wantedInit);
       // Initial configure
--- a/src/app/components/feed/feed-connection.factory.js
+++ b/src/app/components/feed/feed-connection.factory.js
@@ -64,7 +64,7 @@
         options = options || {};
 
         var media = {videoRecv: false, audioRecv: false};
-        var cfg = {video: true, audio: true};
+        var cfg = {video: true, audio: false};
         if (this.role === "main") {
           if (options.noCamera) {
             media.videoSend = false;
--- a/src/app/components/room/room.service.js
+++ b/src/app/components/room/room.service.js
@@ -149,7 +149,7 @@
             connection.publish({
               success: function() {
                   //start session unmuted?
-                  if(participants < 1){
+                  if(participants < 4){
                       toggleChannel('audio');
                   }
               },
--- a/src/app/components/room/room.service.js
+++ b/src/app/components/room/room.service.js
@@ -13,7 +13,7 @@
 
     RoomService.$inject = ['$q', '$rootScope', '$timeout', 'FeedsService', 'Room',
       'FeedConnection', 'DataChannelService', 'ActionService', 'jhConfig',
-      'ScreenShareService', 'RequestService', 'MuteNotifier'];
+      'ScreenShareService', 'RequestService'];
 
   /**
    * Service to communication with janus room
@@ -22,7 +22,7 @@
    */
   function RoomService($q, $rootScope, $timeout, FeedsService, Room,
       FeedConnection, DataChannelService, ActionService, jhConfig,
-      ScreenShareService, RequestService, MuteNotifier) {
+      ScreenShareService, RequestService) {
     this.enter = enter;
     this.leave = leave;
     this.setRoom = setRoom;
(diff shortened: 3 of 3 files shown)


src/app/components/room/room.service.js
```diff
@@ -144,12 +144,21 @@
             ActionService.enterRoom(msg.id, username, connection);
             // Step 3. Establish WebRTC connection with the Janus server
             // Step 4a (parallel with 4b). Publish our feed on server
+
+            var participants = 0;
             connection.publish({
+              success: function() {
+                  //start session unmuted?
+                  if(participants < 4){
```
imobach: What about some kind of "variable" to hold that 4 with a descriptive name?

sandr3j: Sounds useful, if somebody doesn't want to use the default limit. I can add the variable at the beginning of doEnter().

imobach: On the other hand, I think that you can get the list of participants from `msg` here, right? So you won't need that `participants` variable.

sandr3j: I took this approach because msg.publishers instanceof Array could be false, so I have a default value.

imobach: Yes, but what I mean is that you can also check that is not `false` in that callback.

The (rare) problem is that if the callback is really fast and gets executed before the rest of the function, `participants` won't be set at all to the correct value. I'm not sure if such a thing can happen, but...

sandr3j: I had that thought but connection.publish is an async network operation and will never fire that fast, so I wanted to keep the code short.

sandr3j:
based on your feedback I made some modifications

ancorgs:
I like it in general so far. Have you tested it?

sandr3j:
I test everything I commit ;)

imobach:
Ok, so let's merge it. Thanks a lot for your contribution!

src/app/components/feed/connection-config.factory.js
```diff
@@ -24,7 +24,7 @@
     return function(pluginHandle, wantedInit, jsep, ok) {
       var current = {};
       var requested = null;
-      var wanted = {audio: true, video: true};
+      var wanted = {audio: false, video: true};
```
ancorgs: Sure you need to change this IN ADDITION to changing `feed-connection.factory.js`?

sandr3j: It seems like it doesn't make a difference, but with audio: false I get "Connection configured Object {audio: false, video: true}" two times in the console, while with audio: true only one time. When I tested I got some configuration error and thought this was the cause but it isn't. So maybe I should revert this change..

ancorgs:
> I test everything I commit ;)

That's what wanted to read ;)

src/app/components/feed/feed-connection.factory.js
```diff
@@ -64,7 +64,7 @@
         options = options || {};
 
         var media = {videoRecv: false, audioRecv: false};
-        var cfg = {video: true, audio: true};
+        var cfg = {video: true, audio: false};
```
ancorgs: Even more, I probably wouldn't change this value directly here. I would add a new option to `publish`, similar to the already existing `noCamera`. Let's call it, for example, `muted`. And then I would use that argument when calling `publish` (see the other comment).

src/app/components/room/room.service.js
```diff
@@ -144,7 +147,14 @@
             ActionService.enterRoom(msg.id, username, connection);
             // Step 3. Establish WebRTC connection with the Janus server
             // Step 4a (parallel with 4b). Publish our feed on server
+
             connection.publish({
+              success: function() {
```
ancorgs: instead of abusing `success` to activate the audio after having connected muted (quite tricky), I would use the new option explained in the other comment. Something like
```javascript
var startMuted = (msg.publishers instanceof Array) && msg.publishers.length > joinUnmutedLimit;
connection.publish({
    muted: startMuted,
     error: function() { connection.publish({noCamera: true, muted: startMuted}); }
});

File: src/app/components/room/room.service.js [hunk 5]
```diff
@@ -101,6 +101,9 @@
     function doEnter(username) {
       var $$rootScope = $rootScope;
       var connection = null;
+      
+      //max number of participants who will join session unmuted
+      var joinUnmutedLimit = 3;
```
ancorgs: IMO, this should be a configuration option in `config.json`. If set to null (or undef), it would mean "never start muted".

src/app/components/room/room.service.js
ancorgs: It would be nice to explain the user what is happening. Showing a notification "you start muted because bla bla" would be a legit usage for the `success` callback (not saying it's the correct place to put it, but at least it would not be abusing it).

ancorgs:
Now I can say I have done the real code review. I added several comments. To summarize:

- Don't touch `ConnectionConfig`, it's a low level layer that should not be affected.
- Add a new argument to `FeedConnection.publish` instead of changing the default behavior
- Use that new argument instead of abusing the `success` callback
- Move `joinUnmutedLimit` to `config.json`
- (Optional) a notification to the user, so they understand what's going on

PR #127, opened 2016-03-07
Better handling of muting on join

### LINKED ISSUE #72: FR: need push-to-talk mode
Would be useful to be able to configure a keyboard shortcut for push-to-talk like with mumble.

sandr3j:
Based on feedback from @ancorgs I improved automatic muting on join.

### DIFF
--- a/src/app/components/feed/feed-connection.factory.js
+++ b/src/app/components/feed/feed-connection.factory.js
@@ -64,8 +64,11 @@
         options = options || {};
 
         var media = {videoRecv: false, audioRecv: false};
-        var cfg = {video: true, audio: false};
+        var cfg = {video: true, audio: true};
         if (this.role === "main") {
+		  if (options.muted){
+			cfg.audio = false;
+		  }
           if (options.noCamera) {
             media.videoSend = false;
             cfg.video = false;
--- a/src/app/components/feed/feed-connection.factory.js
+++ b/src/app/components/feed/feed-connection.factory.js
@@ -53,12 +53,12 @@
 
       /**
        * Negotiates WebRTC by creating a webRTC offer for sharing the audio and
-       * (optionally) video with the janus server. On success (the stream is
-       * created and accepted), publishes the corresponding feed on the janus
-       * server.
+       * (optionally) video with the janus server. The audio is optionally muted.
+       * On success (the stream is created and accepted), publishes the corresponding
+       * feed on the janus server.
        *
-       * @param {object} options - object with the noCamera boolean flag and
-       *        some callbacks (success, error)
+       * @param {object} options - object with the noCamera boolean flag, muted boolean flag,
+       * and some callbacks (success, error)
        */
       this.publish = function(options) {
         options = options || {};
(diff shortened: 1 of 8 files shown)


src/app/components/feed/feed-connection.factory.js
```diff
@@ -64,8 +64,11 @@
         options = options || {};
 
         var media = {videoRecv: false, audioRecv: false};
-        var cfg = {video: true, audio: false};
+        var cfg = {video: true, audio: true};
         if (this.role === "main") {
+          if (options.muted){
```
ancorgs: Please document the new argument in the heading of the function.

src/app/components/notifier/mute-notifier.service.js
```diff
@@ -27,6 +30,16 @@
       info("Trying to say something? You are muted.");
     }
 
+    function joinedMuted(){
+      info("You're muted. Unmute yourself to join the conversation!");
```
ancorgs: I would like something more explicit about the reason (and we don't need the "unmute yourself" part because we already have a button saying "unmute"). Something like this (just my first attempt SURE you can improve it):
"Joining to a conversation with more than X participants. You'll start muted."

ancorgs:
Wow! Quite impressive but please open separate pull requests for separate features/fixes.

I will only comment here stuff from the start muted feature.

The commit to fix the newVal problem should disappear (we already have a pull request for it)

The commits for push to talk should be moved to a different pull request so we can have a focused conversation there.

Thanks!

src/app/components/feed/jh-feed.directive.js
```diff
@@ -61,6 +61,10 @@
             // Reset the warning timeout
             mutedWarningTimeout = now();
           });
+          scope.$on('muted.Join', function() {
+            mutedWarningTimeout = secondsFromNow(60);
```
ancorgs: Wouldn't it be better to set it to `now()` or to `secondsFromNow(3)` ? It would have a similar effect but with more consistent times (the same time than muting myself or the same one that having been muted). Isn't it?

ancorgs:
Parts about starting muted already commented. Only some nitpicking details here and there, but in general it looks really nice. Great job!

Waiting for a clean pull request to merge that part and release a new version.

src/app/components/feed/jh-feed.directive.js
sandr3j: The reasoning behind this was to stop a second notification (when voice is detected) from appearing while the first is not dismissed yet. But now I see that it didn't fix it properly so I will take another approach..

src/app/components/notifier/mute-notifier.service.js
sandr3j: Yes I thought about "You're muted because there are more than X participants in this room. Unmute yourself..." but I wanted to use a brief sentence to avoid wrapping of the notification on small screens. I will try to make a better sentence.

ancorgs: Yes, wording is usually much harder than coding. :smile:

sandr3j:
Thanks for the kind words, great that you like it :)
I added now new commits and cleaned up this pull request. I will make new pull request for push-to-talk

ancorgs:
Thanks for the changes. It looks nice now. Let's merge.

PR #128, opened 2016-03-08
Add Push-To-Talk feature. This closes #72.

### LINKED ISSUE #72: FR: need push-to-talk mode

sandr3j:
I implemented the requested Push-To-Talk feature. It was necessary to patch a library mousetrap-record licensed under Apache 2.0 license. It was also necessary to use an other angular-hotkey repository with a patched bug.


src/lib/mousetrap-record.js
```diff
@@ -0,0 +1,225 @@
+/**
+ * This extension allows you to record a sequence using Mousetrap.
```
ancorgs: Mmm, I wonder if there is not a better way to include this file.

src/lib/mousetrap-record.js
```diff
@@ -0,0 +1,225 @@
...
+            return;
+        }
+
+        // remember this character if we're currently recording a sequence
+        if (e.type == 'keydown') {
+            if (character.length === 1 && _recordedCharacterKey) {
+                _recordCurrentCombo();
+            }
+
+            for (i = 0; i < modifiers.length; ++i) {
```
ancorgs: This throws an error. There is a missing `var i;` at the beginning of the function.

ancorgs:
Even after fixing the mentioned lack of declaration of `i` in `_handleKey`, the behavior looks buggy to me. Playing a little bit with it, turns out to be relatively easy to reach a situation in which it does not record the last pressed key, but the previous one.

src/app/components/feed/jh-pushtotalk-button.html
```diff
@@ -0,0 +1,10 @@
+<button ng-class="{'btn-danger': vm.hotkeyActive }" class="btn" ng-click="vm.click()" title="{{vm.titleText}}">
+  <span class="glyphicon glyphicon-bullhorn" aria-hidden="true">
+  </span>
+  <span ng-show="vm.showHotkey">
+    <kbd style="background-color:#f15a5c">
+        {{vm.hotkey}}
```
ancorgs: Just a nickpicking detail. When the mode is activated and the key is displayed, the header grows a little bit, at least in Firefox.

src/app/components/room/action.service.js
```diff
@@ -131,5 +132,12 @@
         feed.setEnabledChannel(type, true);
       }
     }
+
+    function setMainFeedChannel(type, boolval) {
```
ancorgs: Pretty non-intuitive name for this function. I know that `setEnabledMainFeedChannel` is ridiculous long, but I'm sure we can find a better alternative.

src/app/components/room/room.service.js
```diff
@@ -416,6 +418,16 @@
       ActionService.toggleChannel(type, feed);
     }
 
+    function pushToTalk(keyevent){
```
ancorgs: Please, try to document new functions added to `RoomService`.

ancorgs:
I like how the feature is implemented from the user interaction point of view. We just need to fix a couple of small bugs.

src/lib/mousetrap-record.js
sandr3j: You're right. On Chrome it worked without var i, and for Firefox I forgot to make the change persistent.

src/app/components/feed/jh-pushtotalk-button.html
sandr3j: Ah I didn't notice that. It grows by one or two pixels.

sandr3j: You're right. I did now some stress testing and I see that it records the previous key. I also uncovered a second bug in my function abortKey, I need to add `if(_recordedSequenceCallback !== null)` because if aborting very fast before it finished triggering all callbacks it throws an error.

src/lib/mousetrap-record.js
sandr3j: We could move it to app/ and exclude in gulp or fix all annoying jslint errors. If the library is complete garbage we could use this: https://dmauro.github.io/Keypress/
But it's a much bigger library in comparison to this small Mousetrap plugin.

src/app/components/room/action.service.js
sandr3j: `setMedia` or `enableMedia` sounds for me more intuitive.

src/app/components/room/room.service.js
sandr3j: Yes, a small description like for `sendStatus` sounds useful.

sandr3j:
Yop, the most annoying bug I see is that the previous key can get recorded. I will try to fix that.

sandr3j:
@ancorgs I pushed now bug fixes. Please review :)

src/lib/mousetrap-record.js
ancorgs: My concern was not about the location of the file. I was thinking about including Mousetrap through bower.json and either contributing the fixes to the upstream Mousetrap either maintaining our own fork with the fixes/enhancement. 

Take into account that currently Mousetrap lives here https://github.com/mousetrap-js/mousetrap (not in the original ccampbell repo). They look very welcoming to fixes and improvements.

So in short, don't copy&paste&modify code but collaborate with the maintainers.

sandr3j: The thing is ccampbell repo seems not to be maintained and I simply missed this fresh fork. The best thing they updated [the plugin](https://github.com/mousetrap-js/mousetrap/blob/master/plugins/record/mousetrap-record.js) with the features we need!

They implemented `stopRecord` function and I can pass `recordSequence=false` to remove the delay after the first key is recorded. I will reimplement key recoding using their library and will add their repository to bower.

sandr3j:
@ancorgs Hey, I thought we should complete this feature regardless of GSoC. The mousetrap record plugin in the mousetrap-js repository was still buggy. I fixed it and [commited my fix to mousetrap-js repository](https://github.com/mousetrap-js/mousetrap/commit/8236fbda90e2c50dfbfbe4e3f26258ad3a873968). 
The plugin should work now snappy and bug free. Please review again if everything is ok so this can get merged :)

ancorgs:
The branch has conflicts. Looks like you need to rebase

sandr3j:
I have taken care of the conflict, beautified the code and added preventDefault for hotkey.

ancorgs:
I have very limited internet access these days. I'll check (and surely merge) tomorrow or during weekend.

ancorgs:
Mmm, I'm having some problems testing this. Looks like gulp is not including the mousetrap plugins into vendor.js. Still researching

src/index.html
```diff
@@ -43,6 +43,7 @@
      <link rel="stylesheet" href="bower_components/angular-gridster/dist/angular-gridster.min.css"/>
      <script src="bower_components/javascript-detect-element-resize/jquery.resize.js"></script>
      <script src="bower_components/angular-gridster/dist/angular-gridster.min.js"></script>
+     <script src="bower_components/mousetrap/plugins/record/mousetrap-record.js"></script>
```
ancorgs: This approach only works with `gulp serve` but not with `gulp build`, so it's not valid.

The three preceding lines (including angular-gridster and jquery.resize) are actually garbage and should also be removed because that code is already included in vendor.js and vendor.css.

We need a clean way to make sure mousetrap-record is included in vendor.js as well.

sandr3j:
Hey, cool that you finally could review my commit! Sorry, I didn't want to cause any headache. I only tested using `gulp serve`. It was very tempting to include it that way and there are even [blogs](http://zellwk.com/blog/bower/#using-a-library-installed-with-bower) which recommend this. Anyway it seems like [wiredep bower overrides](https://github.com/taptapship/wiredep#bower-overrides) is the way to go here. I tested now `gulp build` and there are no problems anymore. I also removed the unused garbage. Please check if this issue is fixed.

ancorgs:
It works nicely now at a first sight. But it seems to have some small issues according to the comments on https://github.com/jangouts/jangouts/issues/72. I also agree with those comments that the setting should be remembered, pretty much like we already do with the layout, the room, the nickname and so on.

This feature is taking quite some time to shape up, but I really really like the result so far. :smile:

sandr3j:
Somehow I spent on this more time then I want to admit. So long that I wrote javascript which freezes Firefox and crashes Chrome. I created an infinite recursive call and Chrome began leaking massive memory until the tab crashed (which I confirmed with gdb). It looks like it only happens while a video feed is streamed. The same code freezes Firefox. But isolating the issue and document it on the bug tracker is not that fun :S

Anyway there is the workaround for the superkey for Firefox and Chrome. And then I uncovered another bug for key combinations. When you release pressed keys in the wrong order the keyup event is fired so I wrote for this a workaround too. Somehow the workarounds are too big and hackish so we might consider to remove support for key combinations in favor of better maintainability. Somehow I didn't expect that these hotkey libraries are so buggy...

ancorgs:
I know that feeling. :(

Just remove support and move forward.

I don't think a "restricted support" message is needed either.  Simply document somewhere in the code the reasons that leaded to the decision. At some point (not as part of this PR) we should create a "help" option (maybe improving the current "keyword shortcuts" one) and there we can explain something to the user. But as said, not now.

sandr3j:
I removed now support for key combinations. Let me know if there is something else that need to be fixed.

src/app/components/feed/jh-pushtotalk.directive.js
```diff
@@ -0,0 +1,148 @@
...
+    return {
+      restrict: 'E',
+      templateUrl: 'app/components/feed/jh-pushtotalk-button.html',
+      controllerAs: 'vm',
+      controller: jhPushToTalkButtonCtrl,
+      scope: {}
+    };
+
+    function jhPushToTalkButtonCtrl() {
+      var titleText1 = "Set Push-to-talk hotkey or combination";
```
ancorgs: It still says "or combination" here.

src/app/components/feed/jh-pushtotalk.directive.js
```diff
@@ -0,0 +1,148 @@
...
+            combo: key,
+            description: 'Push-to-talk',
+            callback: pttCallback,
+            action: 'keyup'
+          });
+
+          vm.hotkey = key;
+          vm.showHotkey = true;
+          vm.titleText = titleText2;
+          UserService.setSetting("lastHotkey", key);
```
ancorgs: This persists the key when push-to-talk is set. But it does not "forget" the setting when push-to-talk is disabled. That's unexpected. If I enable push-to-talk once, I will have it automatically enabled on every subsequent login, even if I disable it every time.

src/app/components/feed/jh-pushtotalk-button.html
```diff
@@ -0,0 +1,10 @@
+<button ng-class="{'btn-danger': vm.hotkeyActive }" class="btn" ng-click="vm.click($event)" title="{{vm.titleText}}">
+  <span class="glyphicon glyphicon-bullhorn" aria-hidden="true">
+  </span>
+  <span ng-show="vm.showHotkey">
+    <kbd style="background-color:#f15a5c;line-height:16px;">
```
ancorgs: This still produces de 1px jump in Firefox. Better use `line-height:1em`, which works in Chrom(e/ium) and Firefox.

sandr3j: Actually I thought the "in the other news"  phrase was just a joke because I don't have this problem  using Firefox 45.

sandr3j:
I fixed now what you mentioned.

ancorgs:
Great job! Thanks a lot!

PR #129, opened 2016-03-12
Fix off-by-one error in startMuted

ancorgs:
This fix a off-by-one error in the feature recently implemented by @sandr3j in #125. The start muted feature was actually activated one participant later than expected. For example, a value of 0 for `joinUnmutedLimit`shows the message "everybody start muted", but actually only the second (and upcoming) participants started muted.

### DIFF
--- a/src/app/components/notifier/mute-notifier.service.js
+++ b/src/app/components/notifier/mute-notifier.service.js
@@ -33,10 +33,10 @@
 
     function joinedMuted(){
       var notiftext = "You joined muted because ";
-      if(jhConfig.joinUnmutedLimit == 0){
+      if(jhConfig.joinUnmutedLimit === 0){
         notiftext += "everyone who joins is muted by default.";
       }
-      else if(jhConfig.joinUnmutedLimit == 1){
+      else if(jhConfig.joinUnmutedLimit === 1){
         notiftext +=  "there is more than one participant.";
       }
       else{
--- a/src/app/components/notifier/mute-notifier.service.js
+++ b/src/app/components/notifier/mute-notifier.service.js
@@ -32,15 +32,15 @@
     }
 
     function joinedMuted(){
-      var notiftext = "You joined muted because ";
+      var notiftext = "You are muted because ";
       if(jhConfig.joinUnmutedLimit === 0){
-        notiftext += "everyone who joins is muted by default.";
+        notiftext += "everyone who enters a room is muted by default.";
       }
       else if(jhConfig.joinUnmutedLimit === 1){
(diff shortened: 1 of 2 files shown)


sandr3j:
Actually I thought it works as intended. I thought that the first one is someone who starts a conversation and the second participant and the others 'join' the conversation. My reasoning was there is no point for the first one to be muted because there is no one who can be disturbed.

ancorgs:
Then the texts are misleading in my opinion. Currently it works like this:

- Limit set to 0. First user joins, not muted. Second participant joins, she starts muted a receives a message saying "everyone who joins is muted by default"... which is not true. In fact, there is no way to enforce that everyone starts muted by default (not discussing if it's useful or not).
- Limit set to 1. First two users join, not muted. Third participant joins muted and receives the message "muted because there is more than one participant"... which is at least confusing.
- Limit set to 2. When the fourth participant joins, she starts muted with a message saying "there are more than 2 participants". Also not 100% clear, in my opinion.

But maybe I'm the only one who finds it confusing. Let's get a second opinion.  @imobach? @ChrisBr?

This pull request "moves" the limits by one, i.e. a limit of 0 actually implies that even the first user starts muted which, in my opinion, better fits the messaging (and probably has some use case).

sandr3j:
My version is:
- Limit set to 0. First user _creates a room / starts the conversation_, not muted. Second participant _joins_, she starts muted and receives a message saying "_everyone who joins_ is muted by default"... Correct, the second participant is the first one who joins or can join because for the first one there is no conversation to join, he has to create / start one.. 
It's correct there's no way to enforce that the starter of the conversation is muted but as I said I thought it's not necessary.
- Limit set to 1. First two users, the conversation starter and the first who joined him, not muted. Third participant joins muted and receives the message "muted because there is more than one participant"... Before the third participant joined, there were 2 participants. 2 participants, means more than one. The third participant is muted because there was already more than one participant in the conversation before he joined. The third participants joins the conversation and is notified that he joined muted because there is more than one participant.

I don't know what is more intuitive. I see that it might be confusing. If we want to add the use case that the first participant, the conversation starter, can start muted we need to make this change.

sandr3j:
I suggest to make this change and change the wording to:

You are muted because ...

- everyone who enters the room is muted by default
- there is more than one participant in the room.
- there are more than X participants in the room.

"The amount of noise generated by a change is inversely proportional to the complexity of the change."

ancorgs:
I like the suggested change. Removing the verb "join" seems to be a good way of avoiding the misunderstanding. I will incorporate those changes.

PR #133, opened 2016-04-03
Allow name change. This closes #15.

### LINKED ISSUE #15: Allow name change
Allow the users to change their name during the session

PaliwalSparsh:
Fixes [#15](https://github.com/jangouts/jangouts/issues/15)
Allows name change of the publisher's feed and the change is reflected in peers feed also.
When publisher double clicks the name in his feed it becomes editable.

Bonus Track : Also the change is reflected in the url. 
Did not add for animation of the feed while changing name (Extra bonus track).

### DIFF
--- a/src/app/components/feed/feeds.factory.js
+++ b/src/app/components/feed/feeds.factory.js
@@ -293,6 +293,30 @@
       };
 
       /**
+       * notifying changes to the remote peers.
+       */
+      this.updateDisplay = function() {
+        var that = this;
+        $timeout(function() {
+          DataChannelService.sendStatus(that);
+        });
+      };
+
+      /**
+       * Sets display attribute for this feed. Works only for remote ones.
+       */
+      this.setDisplay = function(val) {
+        this.display = val;
+      }; 
+
+      /**
+       * Gets the display value for this feed
+       */
+      this.getDisplay = function() {
+        return this.display;
+      };
+
+      /**
        * Reads the representation of the local feed in order to send it to the
        * remote peers.
        *
@@ -304,7 +328,7 @@
         options = options || {};
         if (!options.exclude) { options.exclude = []; }
 
-        var attrs = ["audioEnabled", "videoEnabled", "speaking", "picture"];
+        var attrs = ["audioEnabled", "videoEnabled", "speaking", "picture", "display"];
         var status = {};
 
         _.forEach(attrs, function(attr) {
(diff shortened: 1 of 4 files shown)


src/app/components/feed/jh-feed.html
```diff
@@ -39,7 +39,11 @@
     </div>
   </div>
   <figcaption>
-    <span class="caption">{{ vm.feed.display }}</span>
+    <span class="caption" ng-dblclick="vm.isEditable()" ng-hide="vm.nameIsEditable">{{vm.feed.display}}</span>
```
PaliwalSparsh: on double click name becomes editable span disappears and the div appears, on clicking OK span again appears.

src/app/components/feed/feeds.factory.js
```diff
@@ -293,6 +293,30 @@
       };
 
       /**
+       * notifying changes to the remote peers.
```
PaliwalSparsh: updateDisplay basically sends the status as soon as OK is clicked in the UI, making it as updateLocalPic did not do the trick as there was nothing to update name(display attribute) on the peers side, thus added getDisplay and setDisplay and added display to setStatus to make it work. Please review and correct if this method does not follow jangouts practices.

src/app/components/feed/jh-feed.directive.js
```diff
@@ -116,6 +119,20 @@
         }
       }
 
+      function isEditable(){
```
ancorgs: The name of this function is quite confusing. Functions called `isWhatever` are expected to be checks (with no effect) returning a boolean value.

I would try something closer to `makeNameEditable` (just a possibility).

ancorgs:
From the code point of view, the pull request is quite ok (just a minor problem with a non-intuitive function name).

From the usability point of view, it probably needs some more love. Some things I find inconvenient:

* Some problems with the focus after entering in editing mode. It's very easy to end up pressing backspace and going back in the browser history.
* If the alphabetical order is affected, the position of the feed in the list changes already while writing the name (before pressing 'ok')
* There is no way to cancel the edition (other than pressing 'ok').

Maybe we should go a little bit further in rethinking the UI to accommodate this feature.

sandr3j:
Another bug is that you can use usernames with infinite length or zero length. When I paste a large text as username it will give errors on the other side.

imobach:
Hi @PaliwalSparsh,

Any update about this PR?

Thanks!

imobachgs:
It looks like nobody is interested in working on this PR anymore. Moreover @magarcia is working on the Angular2 version of Jangouts, so I'm closing this PR.

PR #135, opened 2016-04-30
Click on thumbnail to pin webcam

### LINKED ISSUE #131: click anywhere in webcam thumbnail to pin webcam
Currently each thumbnail has a small "Sticky as main video" icon, but this violates principles of good UI based on [Fitts's law](https://en.wikipedia.org/wiki/Fitts's_law).  Instead it would be better to ditch that icon and allow clicking anywhere on the thumbnail to have the same effect (with the obvious exception of clicking on the Ignore icon).

magarcia:
- Add click to thumbnail to pin webcam
- Remove jhStickyButton directive
- Move video button to figcaption, below thumbnail

Closes #131


imobach:
It looks good and you have removed quite some code :) Well done. I want to give it a try in my laptop before merging :)

imobach:
LGTM. We should find another way to arrange mute/video buttons (as they now share the space with the nickname). But from my POV, that's something we should do about the whole interface.

PR #136, opened 2016-04-30
Unpublish shared window/screen feed when stop sharing

### LINKED ISSUE #25: When a shared window is closed, it stays as connected

magarcia:
This solution uses MediaStream onended callback which is [not implemented in Firefox yet](https://bugzilla.mozilla.org/show_bug.cgi?id=1045810).

Tested with Chromium.

Closes #25

### DIFF
--- a/src/app/components/room/room.service.js
+++ b/src/app/components/room/room.service.js
@@ -369,6 +369,12 @@
           console.log(" ::: Got the screen stream :::");
           var feed = FeedsService.find(id);
           feed.setStream(stream);
+
+          // Unpublish feed when screen sharing stops
+          stream.onended = function () {
+            unPublishFeed(id);
+          };
+
         },
         onmessage: function(msg, jsep) {
           console.log(" ::: Got a message (screen) :::");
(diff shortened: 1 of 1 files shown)


magarcia:
Looking inside janus.js code I saw that [they implement](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L1449-L1466) the workaround mentioned in [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1045810), so this should works with Firefox too.

magarcia:
After found and solves an error with `janus.js` (meetecho/janus-gateway#513), I tested with Firefox and it works.

imobach:
Thanks for fixing it! I guess that we also need to update `janus-gateway` dependency in our `bower.json` file, right?

magarcia:
@imobach update `janus-gateway` is not needed for this PR. The issue mentioned above does not affect to these changes.

But we will need to change some code to adapt to changes introduced in the `janus-gateway` update.

magarcia:
I opened an issue about that #139

imobach:
I meant that if we want this bug solved in Firefox, we'll need to update the `janus.js` we're shipping with jangouts. I'm merging this code and looking forward to #139 (as it looks like some additional work is needed).

PR #137, opened 2016-05-05
Vagrant uses openSUSE Leap 42.1 instead of 13.2

imobach:
Vagrant will use a openSUSE Leap 42.1 box.


imobach:
It works for me(TM) :)

Issue #44, opened 2015-08-20
Add a test suite


imobachgs:
Jangouts is getting bigger and bigger so, IMHO, we must build a test suite.

ancorgs:
In my opinion we should have separate issues for unit tests (should be covered by #156) and end-to-end (integration) tests. I would aim for decent unit test coverage for milestone 0.5.0... probably leaving e2e tests for the next version.

imobachgs:
We already have a test suite in the angular-2 branch (only unit tests).

ancorgs:
We have some unit tests now, as part of the migration to React done in version 0.6.0.

It can of course be expanded over time, but I'm closing this because we indeed have a test suite now.

PR #138, opened 2016-05-07
Add test runner and tests for action.service.js

magarcia:
First steps for build a test suite

Related with #44


tests/components/room/action.service.spec.js
```diff
@@ -0,0 +1,665 @@
...
+  });
+
+
+  // Test enterRoom method
+  describe('enterRoom', function () {
+    var feedId = 1,
+        display = 'd',
+        connection = 'conn';
+
+    it('should call Feed',function () {
```
ancorgs: I would like to see more "human" description like "should create the feed"

tests/components/room/action.service.spec.js
```diff
@@ -0,0 +1,665 @@
...
+        jasmine.objectContaining({fakeFeed: true}),
+        jasmine.objectContaining({main: true})
+      );
+    });
+
+  });
+  // End enterRoom method
+
+  // Test leaveRoom method
+  describe('leaveRoom', function () {
```
ancorgs: Let's see if I can explain what I don't like about this "describe" section. It looks to me like it's testing the implementation instead of the behavior. I know it's hard to avoid when functions are that small, but:
```
 leaveRoom
    should call FeedsService.allFeeds
    should call ActionService.destroyFeed for each returned feed
```
Simply looks like 1:1 repetition of the source code.

As said, it's hard to see the difference in such a small function, but I would prefer something like
```
 leaveRoom
    should destroy all existing feeds
```
The implementation of the spec will be almost identical if you mock `FeedService`, but still there is an important difference in the approach.

tests/components/room/action.service.spec.js
```diff
@@ -0,0 +1,665 @@
...
+      var feeds = FeedsServiceSpy.allFeeds();
+      expect(ActionService.destroyFeed.calls.count()).toBe(feeds.length);
+    });
+
+
+  });
+  // End leaveRoom method
+
+  // Test publishScreen method
+  describe('publishScreen', function () {
```
ancorgs: This case is even worse than above. The 1:1 repetition between implementation and test is taken to the extreme. I will try to come up with a working re-implementation of some of the tests in a more behavior-oriented fashion (also as an exercise to learn jasmine myself).

ancorgs:
Ahem, I don't know how I ended up commenting on this old pull request. Too many open tabs. Sorry :wink:

Now let's review the REAL tests that you were adding in the CORRECT pull request :smile:

imobach:
Can we close this one, please?

ancorgs:
I was going to suggest the same.

PR #143, opened 2016-06-10
Angular2

magarcia:
- Migrate Chat component to Angular 2
- Add tests for Chat component
- Apply new application structure to Footer component


config/karma.conf.js
```diff
@@ -87,7 +87,7 @@ module.exports = function(config) {
      * Continuous Integration mode
      * if true, Karma captures browsers, runs the tests and exits
      */
-    singleRun: true
+    singleRun: false
```
ancorgs: I wouldn't change this. In my opinion, `singleRun: true` is better if using vagrant or any similar VM approach because it prints the output to stdout. People wanting CI can always change it.

magarcia: I updated the `package.json` setting the `--single-run` option when execute `npm run test`. I have to do this because setting `singleRun: true` in the karma config don't allows to use the `--auto-watch` option.

src/app/chat/chat-form/chat-form.component.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    let field: string = "text";
+    spyOn(mockActionService, "writeChatMessage");
+
+    (<Control>chatForm.chatForm.controls[field]).updateValue("manual input");
+    chatForm.submit();
+
+    expect(mockActionService.writeChatMessage).toHaveBeenCalledWith("manual input");
+  }));
+
+});
```
ancorgs: For this whole file I would like to suggest this alternative
```typescript
import {
  beforeEach, 
  beforeEachProviders,
  describe,
  expect,
  inject,
  it
} from "@angular/core/testing";

import { Control } from "@angular/common";
    
declare const jasmine;
declare const spyOn;
  
import { ChatFormComponent } from "./chat-form.component";

class MockActionService {
  public writeChatMessage(text: string): void { }
}

describe("ChatForm", () => {
  beforeEachProviders(() => {
    this.actionService = new MockActionService();

    return [
      {provide: ChatFormComponent, useClass: ChatFormComponent},
      {provide: "ActionService", useValue: this.actionService}
    ];
  });

  beforeEach(inject([ ChatFormComponent ], (chatForm)  => {
    this.chatForm = chatForm;
    this.ctrlGroup = chatForm.chatForm.value;
    this.textField = <Control>this.chatForm.chatForm.controls["text"]
  }));

  it("should start with an empty text", () => {
    expect(this.ctrlGroup.text).toBe(null);
    expect(this.textField.value).toBe(null);
  });

  it("should reset text value on submit", () => {
    this.textField.updateValue("manual input");
    this.chatForm.submit();

    expect(this.ctrlGroup.text).toBe(null);
    expect(this.textField.value).toBe(null);
  });

  it("should use ActionService.writeChatMessage to send the message", () => {
    spyOn(this.actionService, "writeChatMessage");

    this.textField.updateValue("manual input");
    this.chatForm.submit();

    expect(this.actionService.writeChatMessage).toHaveBeenCalledWith("manual input");
  });
});
```

Main changes:

 * Declare the injection only once, using `beforeEach`
 * Using the `this` object provided by Jasmine (a.k.a. `userContext`) not only to DRY the injection but also to have reusable meaningful names.
 * Focus in the behavior, not in the implementation. That means fewer tests with high-level descriptive (hopefully) texts.

In general, I find my version to be more readable. @magarcia, @imobachgs What do you guys think?

src/app/index.ts
```diff
@@ -5,7 +5,9 @@
  * of the MIT license.  See the LICENSE.txt file for details.
  */
 
-import "zone.js";
+//import "zone.js";
```
ancorgs: We should add a comment here explaining the reason for this (explicitly including part of zone.js instead of the whole thing).

DEVELOPMENT.md
```diff
@@ -133,28 +133,21 @@ npm run build:prod
...
-npm run webdriver:start
+npm run test:watch
 ```
 
-### run Protractor's elementExplorer (for end-to-end)
-```bash
-npm run webdriver:start
-# in another terminal
-npm run e2e:live
-```
```
ancorgs: The documentation is more clear now without surplus information. Thanks.

src/app/chat/chat-form/chat-form.component.spec.ts
magarcia: @ancorgs I find it more readable too. I will update the test with your proposal and change some other test I have in other branch to fit this style before the next pull request.

PR #144, opened 2016-06-28
mention required proxy module

bmwiedemann:
without this, apache complains
```
Invalid command 'ProxyVia', perhaps misspelled or defined by a module not included in the server configuration
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,8 @@ system would be:
 
 1. `sudo zypper in apache2`
 2. Copy the content of `dist` directly into `/srv/www/htdocs/`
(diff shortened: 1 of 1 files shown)


ancorgs:
Sure? AFAIR, those instructions are for the quick&dirty way, in which the browser accesses janus directly, not using Apache as proxy. This method uses default Apache configuration (the instructions even suggest to copy directly to `/srv/www/htdocs/`so it's not even needed to configure the document root).

bmwiedemann:
ah, indeed. Without the copying, the grant would be missing.

PR #147, opened 2016-07-06
Angular2 feed

magarcia:


ancorgs:
This pull request looks quite messy. Some examples:

It includes commit 4ea0954fa4a4bf172ef90a9ca29debd7ee9257b5 which is already in the Angular-2 branch. Even worst. The dummy test that commit is supposed to delete is present in the result of the PR.

Something similar happens with many other commits (like 6227e1883b60d99eda858203ba582da584cdf65e), that are already in our angular-2 branch but are also included in this PR.

The PR adds 6 merges to the history. It's an incremental change on what we currently have in angular-2. It should apply cleanly without all those merges.

Please, try to use rebase, cherry-pick and friends to build a clean PR. Ask for help if you need it.

magarcia:
@ancorgs sorry, caos with branches. I think this PR is most clear now.

src/app/feed/buttons/audio-button.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  public showsEnable(): boolean {
+    return (this.feed.isPublisher && !this.feed.isLocalScreen && !this.feed.getAudioEnabled());
+  }
+
+  public showsDisable() {
+    return (!this.feed.isIgnored && this.feed.getAudioEnabled());
+  }
+
+  public showsAudioOff() {
+    return (!this.feed.isPublisher && !this.feed.isIgnored && !this.feed.getAudioEnabled());
```
imobachgs: You could apply the De Morgan's law here:

```!(this.feed.isPublisher || this.feed.isIgnored || this.feed.getAudioEnabled())```

File: src/app/feed/shared/connection-config.model.ts [hunk 2]
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright (C) 2015 SUSE Linux
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE.txt file for details.
+ */
+
+import * as _ from "lodash";
```
imobachgs: Maybe `import _ from 'lodash';` would work.

src/app/feed/shared/connection-config.model.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export interface IWanted {
+  audio?: boolean;
```
imobachgs: Do we need them to be optional? After all, they're just booleans.

src/app/feed/shared/feed-connection.model.ts
```diff
@@ -0,0 +1,194 @@
...
+   *                * values: object with the wanted values for the flags
+   *                * ok: callback to execute on confirmation from Janus
+   */
+  public setConfig(options: IOptions): void {
+    if (this.config) {
+      this.config.set(options);
+    } else {
+      this.config = new ConnectionConfig(this.pluginHandle, options.values, null, options.ok);
+    }
+  };
```
imobachgs: Remove the semicolon, please.

src/app/feed/shared/feed.model.ts
```diff
@@ -0,0 +1,337 @@
...
+  private videoRemoteEnabled: boolean = true;
+  private audioRemoteEnabled: boolean = true;
+  private stream: any = null;
+  private speakObserver: any = null;
+  private dataChannel: any = null;
+
+  /*
+   * Constructor should only require injected parameters, this is the reason to
+   * has setAttrs method. When full app upgraded to Angular 2 DataChannel service
+   * should be injected in the consturctor.
```
imobachgs: typo in *constructor*.

src/app/feed/shared/connection-config.model.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      config,
+      {},
+      undefined
+    );
+
+    expect(cconfig).toBeDefined();
+    expect(this.pluginHandle.send).toHaveBeenCalled();
+  });
+
+  it("should not be defined until confirm", () => {
```
imobachgs: When you write unit tests, you're testing/describing the public API. So I would use `describe` to make clear what you're testing. For example, here you're testing the `get` method.

```javascript
describe("#get", () => {
  it("...")
});
```

On the other hand, with the current test, it's not clear what's expected from the `get` method.

src/app/feed/shared/connection-config.model.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    spyOn(window.console, "error");
+
+    expect(cconfig.get()).not.toBeDefined();
+
+    cconfig.confirm();
+
+    expect(window.console.error).toHaveBeenCalled();
+  });
+
+  it("should call ok callback when confirm", () => {
```
imobachgs: In this case, for example, you're testing the `confirm` method. So just wrap it in a describe.

src/app/feed/shared/connection-config.model.ts
magarcia: For some reason `import _ from "lodash";` not works correctly, but using `import * as _ from "lodash";` solves the problem.

src/app/feed/shared/connection-config.model.ts
magarcia: In [this method](https://github.com/jangouts/jangouts/pull/147/files#diff-b5c6bf8f85bc96337664b5bdf08b7ac6R100) the requested attribute is set as an empty object, so audio and video need to be optional in the interface. I will put a TODO comment to review this code, but by now I want to keep moving with the migration.

magarcia:
I updated the tests with your suggestions and added some missing tests for concrete methods in services and components that had not tested. The `feed/buttons` tests still missing, but I'm on it.

src/app/feed/feed.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    };
+    this.feedComponent.highlighted = false;
+    this.feedComponent.highlightedByUser = false;
+
+    this.feedComponent.ngOnInit();
+  }));
+
+  describe("#thumbnailTag", () => {
+
+    it("should return placeholder if highlighted", () => {
```
imobachgs: That's not important (just a note), but you can get rid of the _should_ to make the description shorter: "returns placeholder if highlighted".

src/app/feed/shared/connection-config.model.spec.ts
```diff
@@ -41,111 +42,97 @@ describe("Service: ConnectionConfig", () => {
...
-    let cconfig: ConnectionConfig = new ConnectionConfig(
-      this.pluginHandle,
-      config,
-      {},
-      undefined
-    );
-    spyOn(window.console, "error");
-
-    expect(cconfig.get()).not.toBeDefined();
+    it("should show an error when pluginHandle raise an error", () => {
```
imobachgs: raise -> raises

src/app/feed/shared/connection-config.model.ts
```diff
@@ -7,6 +7,7 @@
 
 import * as _ from "lodash";
 
+// [TODO] - Review if attrs should be optionals (see `configure` method)
```
imobachgs: optionals -> optional

src/app/feed/shared/feed-connection.model.spec.ts
```diff
@@ -30,102 +30,248 @@ describe("Service: FeedConnection", () => {
...
+            videoSend: true,
+            data: true
+          },
+          success: jasmine.any(Function),
+          error: jasmine.any(Function)
+        })
+      );
+    });
+
+    it("should call success callback when webRTC offer success", () => {
```
imobachgs: success -> succeed

src/app/feed/shared/feed-connection.model.spec.ts
```diff
@@ -30,102 +30,248 @@ describe("Service: FeedConnection", () => {
...
+      this.pluginHandle.createOffer.and.callFake((options) => {
+        options.success()
+      });
+      this.connection.publish(this.options);
+
+      expect(this.options.success).toHaveBeenCalled();
+    });
+
+    it("should call success callback when webRTC offer success", () => {
```
imobachgs: success -> succeeds

src/app/feed/shared/feed-connection.model.spec.ts
imobachgs: I think this test does not belong to here as you're testing the internals of plugin handle. IMHO it makes sense as a pluginHandle unit test, but not as part of `FeedConnection` model.

On the other hand, you're faking the `createOffer` (and calling `options.success()`), so I think that the original code does not get executed.

The same applies to the test below.

src/app/feed/shared/feed-connection.model.spec.ts
```diff
@@ -30,102 +30,248 @@ describe("Service: FeedConnection", () => {
...
+            videoSend: false,
+            data: true
+          },
+          success: jasmine.any(Function),
+          error: jasmine.any(Function)
+        })
+      );
+    });
+
+    it("should send a request start message when webRTC answer success", () => {
```
imobachgs: As above, I think this test does not cover the original code, right? Testing anonymous callbacks in this case (depending on `pluginHandle` and connections) can be tricky. I think that we should try another approach (de-anonymizing the callbacks, for example).

However, I don't want this issue to block us, so I propose postponing these tests (remove the current ones). Please, add a Trello card to not forget about them.

src/app/feed/shared/feed-connection.model.spec.ts
```diff
@@ -30,102 +30,248 @@ describe("Service: FeedConnection", () => {
...
+          message: {
+            request: "start",
+            room: 1
+          },
+          jsep: this.jsepVal
+        })
+      );
+    });
+
+    it("should shows an error when webRTC answer fails", () => {
```
imobachgs: Same as above.

src/app/feed/shared/feed-connection.model.spec.ts
```diff
@@ -30,102 +30,248 @@ describe("Service: FeedConnection", () => {
...
+      this.options = {
+        values: {
+          audio: true,
+          video: true
+        },
+        ok: jasmine.createSpy("okCallback")
+      };
+    });
+
+    it("should create a new config if not exist", () => {
```
imobachgs: if it doesn't exist

src/app/feed/shared/feed-connection.model.spec.ts
```diff
@@ -30,102 +30,248 @@ describe("Service: FeedConnection", () => {
...
+    it("should return undefined if config not exist", () => {
+      let result: any = this.connection.getConfig();
+      expect(result).not.toBeDefined();
+    });
+
+    it("should return the existing config", () => {
+      this.connection.config = jasmine.createSpyObj("config", ["get"]);
+      let result: any = this.connection.getConfig();
+
+      expect(this.connection.config.get).toHaveBeenCalled();
```
imobachgs: You're testing the `config.get` is called but you're not testing whether it's returned. So the following implementation would pass the test:

```javascript
if (this.config) {
  this.config.get();
  return "another-thing";
}
return;
```

src/app/feed/shared/feed.model.spec.ts
```diff
@@ -81,77 +78,70 @@ describe("Service: Feed", () => {
...
-      let feed: Feed = new Feed();
-      feed.setAttrs({
+      this.feed.setAttrs({
         isPublisher: true,
         connection: connection
       });
 
-      let result: boolean = feed.isEnabled("audio");
+      let result: boolean = this.feed.isEnabled("audio");
       expect(connection.getConfig).toHaveBeenCalled();
```
imobachgs: I would just say:

```javascript
expect(this.feed.isEnabled("audio")).toBe(connectionStatus.audio)
```

imobachgs: Moreover you can use describes like this (just an sketch to illustrate the general idea):

```javascript
describe("when audio is enabled in the connection", function() {
  let connectionStatus: any = {
    audio: true
  };

  it("returns true", function() {
    expect(this.feed.isEnabled("audio")).toBe(true);
  });
});

File: src/app/feed/shared/feed.model.spec.ts [hunk 18]
```diff
@@ -194,136 +184,279 @@ describe("Service: Feed", () => {
...
-      let feed: Feed = new Feed();
-      feed.setAttrs({
+      this.feed.setAttrs({
         isPublisher: true,
         connection: connection
       });
 
-      let result: boolean = feed.getVideoEnabled();
+      let result: boolean = this.feed.getVideoEnabled();
       expect(connection.getConfig).toHaveBeenCalled();
```
imobachgs: Implementation detail that you could ignore.

src/app/feed/shared/feed.model.spec.ts
```diff
@@ -194,136 +184,279 @@ describe("Service: Feed", () => {
...
 
-      expect(feed.isDataOpen()).toBe(false);
+      expect(this.feed.isConnected()).toBe(true);
     });
   });
 
-  describe("isConnected", () => {
-    it("should return false if not connection setted", () => {
-      let feed: Feed = new Feed();
+  describe("#igonre", () => {
```
imobachgs: typo: 'ignore'

src/app/feed/shared/feed.model.spec.ts
```diff
@@ -194,136 +184,279 @@ describe("Service: Feed", () => {
...
+    it("should destroy the connection", () => {
+      let connection: any = {
+        destroy: jasmine.createSpy("connection.destroy")
+      };
+      this.feed.setAttrs({ connection: connection });
 
-      expect(feed.isConnected()).toBe(false);
+      this.feed.disconnect();
```
imobachgs: I think you wanted to call `ignore`, right?

src/app/feed/shared/feed.model.spec.ts
```diff
@@ -194,136 +184,279 @@ describe("Service: Feed", () => {
...
-    it("should return true if connection setted", () => {
-      let feed: Feed = new Feed();
-      feed.setAttrs({
-        connection: {}
-      });
+  });
+
+
+  describe("#stopIgnoring", () => {
+    it("should be connected", () => {
```
imobachgs: "should set the feed as connected" or something like that. What do you think?

src/app/feed/shared/feed.model.spec.ts
```diff
@@ -194,136 +184,279 @@ describe("Service: Feed", () => {
...
-      feed.disconnect();
-      feed.ignore();
+  describe("#isSilent", () => {
+    it("should return false if feed is speaking", () => {
+      this.feed.speaking = true;
+      expect(this.feed.isSilent()).toBe(false);
+    });
 
-      expect(feed.waitingForConnection()).toBe(false);
+    it("should return false if feed was speaking in a time given", () => {
```
imobachgs: *given time*

src/app/feed/shared/feed.model.spec.ts
```diff
@@ -194,136 +184,279 @@ describe("Service: Feed", () => {
...
+      this.feed.silentSince = (Date.now() - 1000);
+
+      expect(this.feed.isSilent(5000)).toBe(false);
     });
 
-    it("should return true when connection is not setted", () => {
-      let feed: Feed = new Feed();
-      feed.setAttrs({});
-      feed.disconnect();
+    it("should return true if feed is not speaking in a time given", () => {
```
imobachgs: *given time*

src/app/feed/shared/feeds.service.spec.ts
```diff
@@ -71,8 +74,8 @@ describe("Service: FeedsService", () => {
     });
   });
 
-  describe("allFeeds",  () => {
-    it("should return and empty array when not feeds added", () => {
+  describe("#allFeeds",  () => {
+    it("should return an empty array when not feeds added", () => {
```
imobachgs: *no feeds were added* or *no feeds are present* or...

src/app/feed/shared/feeds.service.spec.ts
```diff
@@ -209,36 +198,72 @@ describe("Service: FeedsService", () => {
...
+      this.feed = new Feed();
+      this.feed.setAttrs({id: 1});
+    });
 
-      this.feeds.waitFor(1, 5, 10);
+    it("should return a promise that resolve with asked feed", <any>fakeAsync((): void => {
+      let response: Promise<any> = this.feeds.waitFor(1, 3, 10);
 
       tick(10);
       tick(10);
```
imobachgs: I know it's not part of this PR but, why do you call `tick` twice?

src/app/feed/shared/feeds.service.spec.ts
```diff
@@ -209,36 +198,72 @@ describe("Service: FeedsService", () => {
...
-      tick(10);
-      (<any>this.feeds.find).and.callFake((): Feed => {
-        let feed: Feed = new Feed();
-        feed.setAttrs({id: 1});
-        return feed;
+
+      expect(this.feeds.find.calls.count()).toBe(4);
+
+      flushMicrotasks(); // resolve promieses
```
imobachgs: typo: *promises*

src/app/feed/shared/feeds.service.spec.ts
```diff
@@ -209,36 +198,72 @@ describe("Service: FeedsService", () => {
...
+      tick(1000);
+      tick(1000);
 
-      expect((<any>this.feeds.find).calls.count()).toBe(6);
+      flushMicrotasks(); // resolve promieses
+
+      expect(rejectHandler).toHaveBeenCalled();
     }));
 
+    it("should return a promise that resolve with the asked feed even if it means not wait", <any>fakeAsync((): void => {
```
imobachgs: *not to wait*

src/app/feed/shared/speak-observer.service.spec.ts
```diff
@@ -34,136 +34,142 @@ describe("Service: SpeakObserver", () => {
...
 
-    expect(window.clearInterval).toHaveBeenCalled();
+      expect(window.clearInterval).toHaveBeenCalled();
+    });
   });
 
-  it("should return not speaking when no sound", <any>fakeAsync((): void => {
-    let speakObserver: SpeakObserver = new SpeakObserver({});
+  describe("#isSpeaking", () => {
+    it("should return not speaking when no sound", <any>fakeAsync((): void => {
```
imobachgs: It returns "false", not "not speaking".

src/app/feed/shared/speak-observer.service.spec.ts
```diff
@@ -34,136 +34,142 @@ describe("Service: SpeakObserver", () => {
...
-      };
-    });
+      speakObserver.destroy();
+    }));
 
-    // run 3 iterations to detect sound
-    tick(65);
-    tick(65);
-    tick(65);
+    it("should return speaking when sound", <any>fakeAsync((): void => {
```
imobachgs: It should return "true".

src/app/feed/shared/speak-observer.service.spec.ts
```diff
@@ -34,136 +34,142 @@ describe("Service: SpeakObserver", () => {
...
+      let speakObserver: SpeakObserver = new SpeakObserver({});
+      spyOn(this.Analyser, "getFloatFrequencyData").and.callFake((fftBins: any) => {
+        for (let i: number = 0; i < 512; i++) {
+          fftBins[i] = -5;
+        };
+      });
 
-    expect(speakObserver.isSpeaking()).toBe(true);
+      // run 3 iterations to detect sound
```
imobachgs: :+1:

src/app/feed/shared/speak-observer.service.spec.ts
```diff
@@ -34,136 +34,142 @@ describe("Service: SpeakObserver", () => {
...
-  }));
+      expect(speakObserver.isSpeaking()).toBe(true);
 
-  it("should call AudioContext.resume when state is suspended", <any>fakeAsync((): void => {
-    let speakObserver: SpeakObserver = new SpeakObserver({});
+      speakObserver.destroy();
+    }));
 
-    tick(65);
+    it("should return not speaking when sound stops", <any>fakeAsync((): void => {
```
imobachgs: "not speaking" => "false"

imobachgs:
In general it looks good. I only found some issues in the tests but I think they are rather easy to solve. Thanks!

src/app/feed/shared/feeds.service.spec.ts
magarcia: I don't know why but if I run `tick(20)` once fails and run `tick(10)` twice works. So for each interval cycle is necessary run one tick.

I will add a note above this code to explain this case.

src/app/feed/shared/feed-connection.model.spec.ts
magarcia: I don't think I'm testing the `pluginHandler`, in these tests the `pluginHandler` it's a fake. In this test I'm faking the `createAnswer` function to test the anonymous callbacks, but this could be a problem if  `pluginHandler` changes his public functions.

De-anonymizing callbacks could be a solution, but IMHO this callback should be private methods and could not be easily testable.

src/app/feed/shared/feed-connection.model.ts
```diff
@@ -104,6 +104,7 @@ export class FeedConnection {
       media.data = false;
     }
 
+    // [TODO] - De-animize the callbacks a tests it
```
imobachgs: *[TODO] de-anonymize callbacks and test them*

src/app/feed/shared/feed-connection.model.ts
```diff
@@ -127,6 +128,8 @@ export class FeedConnection {
    * to a feed from the janus server.
    */
   public subscribe(jsepVal: any): void {
+
+    // [TODO] - De-animize the callbacks a tests it
```
imobachgs: Same as above.

src/app/feed/shared/feed.model.spec.ts
```diff
@@ -240,7 +238,7 @@ describe("Model: Feed", () => {
 
 
   describe("#stopIgnoring", () => {
-    it("should be connected", () => {
+    it("should sets the feed as connected", () => {
```
imobachgs: Sorry! My mistake here: *should set*

imobachgs:
Only two typos. It almost LGTM :)

imobachgs:
LGTM

PR #148, opened 2016-07-07
login-ui-improvement

cyntss:
@ancorgs I need to create the PR now so the branch doesnt get old. But when im back from Argentina I will try to find some time to finish what I was implementing as fixes in the UX, maybe this time with Angular 2.0 once it gets merged :)


cyntss:
@ancorgs can I merge this?

ancorgs:
To be honest, we haven't check. I'll do this weekend. But we should probably keep the UI changes in a different branch until they are more or less complete. Pretty much as we are doing with the angular2 branch.

We make the regular releases out of master so it should be reasonably consistent.

As said, I would like to take a look before merging. :wink:

cyntss:
@ancorgs any news about this PR?
I would like to have the new login merged and then continue working on improvements in the chat room. But go step by step. There are plenty of improvements to make in the general UI to make it look pretty, but wanted to start with the login page first, because to be fair, it was bothering me the ugliness :P 

if everyone else considers we should work on a "release based" branch for UI improvements only, I would be afraid it will only get old and old diverging from master as you make changes everywhere else. But again, up to all of you to decide.

ps: if we go for an approach more CD (continues delivery) I would invest some time during hackweek, but again, i rather have the PR merged as soon as possible to then be able to get feedback from people using it sooner rather than later.

ancorgs:
Let's move this forward.

I still think we should work in a branch different to master. To be honest, master does not move that fast. Specially regarding UI. So doing a `merge` or `rebase` once in a while should not be that hard.

I don't want to mix styles in master because is the "public release" branch, but I'm fine with deploying the new branch in jangouts.suse.de in a regular basis. So we do CD of the new branch on the most used Jangouts instance (the SUSE internal one), without bothering other users (yes, there is people out there using Jangouts).

imobachgs:
@cyntss Hi! I've created a [Hack Week project](https://hackweek.suse.com/15/projects/2122) just in case you're interested. Please, feel free to join :)

@ancorgs @cyntss I think we should start working on top of the `angular-2` branch (including UX stuff). Let's see if we can finish the port during the Hack Week.

cyntss:
@ancorgs ok, I agree with that.

The question now is, should we do what @imobachgs  just recommended, working on the Angular 2 branch? because I would like to work on whatever it is we will have in SUSE to get as much feedback as possible.

Something else, I saw @imobachgs mentioned in the hackweek project that someone recommended using Material Design. I am against that proposition. We made the mistake of trying it in another project, and we really regret it. The documentation is really bad and it is 100% flex based, which is actually not a good practice since Flex should be applied in special cases when needed, not for an absolute griding system. This makes Angular Material really "un-unflexible" for certain UI design
in terms that everything is flexible and that is not always the case.
In the other hand, as far as I know, there is still no documentation for Material design for angular 2
since it is under development.

Last, but not least, I understand the concept from Google of unifying the look and feel of the products people, users in general, use around the internet. But isn't it to their advantage only since we would all be implementing Google style in our products to then lose our identity?
I think that we are already a "clone" of a google product, and now making it look more like Google, would be a terrible mistake. 

We are building (yes, slowly) an identity different to Google Hangout for a reason. To stand out and be something unique, open source, completely free, etc. I would stick to that direction.

Will add my response to the thread as well.

imobachgs:
@cyntss If you want to get early feedback, maybe the `angular-2` branch is not your best option. But, to be honest, I would like to avoid diverging too much. I will try to invest some time before Hack Week starts in order to make the Angular 2 version usable again (we're not that far away) so you can work on top of it.

About Material Design, you've convinced me. We though about it as a quick way of having an improved UI. So let's forget about Material :)

ancorgs:
@cyntss I created a new branch called `new_look` based on the current master. I rebased you pull request and merged it there. From now on, we will always keep that branch up-to-date with `master`. Every commit added to `master` will be merged to `new_look`.

I have deployed the result in https://jangouts.tk/new_look/ and https://jangouts.suse.de/new_look/
I'm fine with deploying it directly to https://jangouts.suse.de/ (without the extra directory) as soon as it's a little bit more polished (right now it's quite hard to read the content of the widgets and the browser-related indications are not there).

Last but not least, please add an entry to https://github.com/jangouts/jangouts/blob/master/LICENSE-artwork.txt for every asset that you use and it's not produced by you. Jangouts is free software and _noblesse oblige_. :-)

cyntss:
Excellent @ancorgs I will work on that branch. I guess you will sort things out with @imobachgs regarding the branches of Angular 1.x and 2

I will add the website I used to get the random picture for the new background ;)

PR #151, opened 2016-08-14
Final migration

magarcia:
# TODO
- Add modal lib [angular2-modal](https://github.com/shlomiassaf/angular2-modal/)
- Add thumbnails mode in signin form
- Fix pushtotalk
- Move to Angular2 routes
- Remove Angular1


src/app/shared/broadcaster.service.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  it("should be notified on event raise", () => {
+    let eventHandler: any = jasmine.createSpy("eventHandler");
+    this.broadcaster.on("MyEvent").subscribe(eventHandler);
+
+    this.broadcaster.broadcast("MyEvent", "example message");
+
+    expect(eventHandler).toHaveBeenCalledWith("example message");
+  });
+
+  it("should be notified only for the event subscribed", () => {
```
ancorgs: A little bit nitpicking, but a more accurate description for this block would be something like
```
it(""should not be notified if not subscribed to the event")
```

src/app/shared/broadcaster.service.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { Injectable } from "@angular/core";
+
+interface IBroadcastEvent {
+  key: any;
+  data?: any;
+}
+
+@Injectable()
+export class Broadcaster {
```
ancorgs: I know this is a pretty trivial class and that we have a separate Trello card to improve doc, but when adding new stuff it would be nice to add at least top-level class documentation to it. Something like (sure you can come up with a better description):

```typescript
/**
 * Class implementing a broadcast mechanism equivalent to the one available in Angular 1
*/

File: src/app/notification/notification.component.ts [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+
+import { Broadcaster } from "../shared";
+import { ActionService } from "../room";
+import { ConfigService } from "../config.provider";
+
+@Component({
+  selector: "jh-notification",
+  template: require("./notification.component.html")
+})
+export class NotificationComponent implements OnInit {
```
ancorgs: Same comment about top-level class documentation

ancorgs: Looks like this class is only used to tell the user he is muted. In that case, `NotificationComponent` looks a little bit generic as a name. Shouldn't be something like `MuteNotificationComponent' (or `MuteAlertComponent` if you want something shorter).

src/app/screen-share/screen-share-help.modal.html
```diff
@@ -0,0 +1,23 @@
+<p>
```
ancorgs: What happened to capital letters in this text?

ancorgs:
I'm reviewing in a commit by commit basis, so expect some outdated comments. :wink:

And about that, when moving files (or turning them into their Angular2 equivalent, to be more precise) it would have been useful to do the `mv` in a separate commit without code changes. So git would have recognized it as the same file despite the format changes.

src/app/feed/buttons/audio-button.component.ts
```diff
@@ -19,14 +20,14 @@ export class AudioButtonComponent implements OnInit {
...
-              @Inject("MuteNotifier") private muteNotifier: any) { }
+              private broadcaster: Broadcaster) { }
 
   public ngOnInit(): void { }
 
   public toggle(): void {
     this.roomService.toggleChannel("audio", this.feed);
     if (this.feed.isPublisher && !this.feed.isLocalScreen && !this.feed.getAudioEnabled()) {
-      this.muteNotifier.dismissLastNotification();
+      this.broadcaster.broadcast("dismissLastNotification");
```
ancorgs: Wow, this change is weird.

src/app/feed/feed.component.scss
```diff
@@ -11,8 +11,8 @@
       cursor: pointer;
       .face {
         vertical-align: middle;
-        width: 64px;
-        height: 48px;
+        width: 128px;
```
ancorgs: Since I guess this change is temporary, we should add a comment so we don't forget.

src/app/index.ts
```diff
@@ -19,21 +18,22 @@ import { upgradeAdapter } from "./adapter";
...
-import config from "./config.provider";
-import signin from "./signin";
+upgradeAdapter.upgradeNg1Provider("hotkeys"); // needed for pushToTalk
 
-// Components
-import footerComponent from "./footer";
-import feedComponent from "./feed";
-import chatComponent from "./chat";
-import roomComponent from "./room";
+/* Register providers for browser, this is mandatory. */
```
ancorgs: I'm a little bit confused by this comment. It claims that something is mandatory... but it's commented.

src/app/videochat/video-chat.component.ts
```diff
@@ -38,13 +39,25 @@ interface IHighlight {
...
+
+
 @Component({
   selector: "jh-video-chat",
   template: require("./video-chat.component.html"),
   directives: [
     ThumbnailsModeButtonComponent,
     MainFeedComponent,
-    PushToTalkComponent,
+    //PushToTalkComponent,
```
ancorgs: When commenting a component, it would be nice to put a FIXME or something. FIXME and TODO are really useful to grep for temporary stuff.

src/app/index.ts
magarcia: This comment comes from the example of angular2-modal, it's not enabled in this branch because needs a full migration to Angular 2. In the next PR removing Angular 1, this should be fixed.

PR #152, opened 2016-08-22
Remove angular1

magarcia:
# Known bugs
- Links in modals doesn't work (need angular2-modal upgrade)
- Speak indicator icon works erratically
- Missing indicator for "sticky" feed
- Push to talk works is weird

# Missing features
- Hotkeys cheat sheet
- Gridster suppport
- Embeded options in chat


ancorgs:
Not this PR's fault, but again I'm having problems to test this because I cannot install typings from most networks. Depending on the network I get different errors but all of them seem to be network related. Typings feel like a very weak thing to me... and according to the number of open issues at Typings github, I'm not the only one. :-(

Anyway, I will comment on the code now and will try to test during the afternoon.

src/app/footer/index.ts
```diff
@@ -1,6 +1,6 @@
-import { upgradeAdapter } from "../adapter";
-import { FooterComponent } from "./footer.component";
+//import { upgradeAdapter } from "../adapter";
+export { FooterComponent } from "./footer.component";
```
ancorgs: Any reason to leave this commented in the final version instead of simply removing it?

src/app/footer/index.ts
```diff
@@ -1,6 +1,6 @@
...
 
-export default angular.module("janusHangouts.footerComponent", [])
-  .directive("jhFooter",
-             <angular.IDirectiveFactory>upgradeAdapter.downgradeNg2Component(FooterComponent));
+//export default angular.module("janusHangouts.footerComponent", [])
```
ancorgs: Same here. Commented code looks like leftover

ancorgs:
I tried again deleting my virtual machine and performing a `vagrant up` from scratch. In that clean environment I'm still unable to install the typings. It fails in various ways. :-(

imobachgs:
I've tried this PR and it works for me (apart from the issues mentioned in the description).

Issue #140, opened 2016-05-23
Placeholders in the configuration URLs


ancorgs:
Some configuration parameters like "janusServer", "janusServerSSL" or "httpsUrl" are URLs. In many situations it would be useful to accept placeholders as part of those urls, specially the host name, so it's possible to specify something like these in servers replying to different hostnames.

``` json
{
  "janusServer": "ws://%{hostname}/janus",
  "janusServerSSL": "wss://%{hostname}/janus"
}
```

The placeholders would be substituted by the corresponding part of the current URL the browser is using to access Jangouts.

mohsini172:
Can I add this using gulp?

mohsini172:
I have modified the gulp serve task and implemented this functionality. I need some guide if I can use the gulp. If yes how the parameter will be passed? Do I need to edit the other tasks for this?

imobachgs:
@mohsini172 Sorry, but the idea is not to use gulp, but to replace those placeholders with information from the URL on runtime. For instance, given the example configuration in this issue, if you're connecting to Jangouts using an URL `https://myserver.com`, then `janusServer` value should be `ws://myserver.com/janus` and it will happen automatically.

Don't hesitate to ask if it's still not clear.

mohsini172:
I'm little confused what exactly issue is? Can you give a practical example to explain this. I mean what will be the initial configuration file will be look like and when run the server from which point in the code should I replace those values?

imobachgs:
@mohsini172 Sure! Given a configuration file with these values:

```json
{
  "janusServer": "ws://%{hostname}/janus",
  "janusServerSSL": "wss://%{hostname}/janus"
}
```

If you connect your browser to `http://myserver.com/`, then `janusServer` should be interpreted as `ws://myserver.com/janus`. Of course, you don't need to change the configuration file itself, but just to modify the configuration on runtime. The benefit is that you won't need to adjust the configuration file by hand to match the hostname.

Is it more clear now?

Thanks!

mohsini172:
Perfectly understand. Just one more thing from which file can I start with. I mean in which file the configuration is reading.

imobachgs:
Closed by #177.

PR #177, opened 2017-03-07
placeholder is added

mohsini172:
#140 index.js reads the configuration and replace with the url

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -76,8 +76,9 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
     if (request.status === 200) {
       var config = JSON.parse(request.responseText);
       angular.forEach(config, function(value, key) {
-         jhConfig[key] = value;
+         jhConfig[key] = value.replace("%{hostname}",window.location.host);
       });
+      console.log(jhConfig);
     } else {
       console.warn('No configuration found');
     }
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -69,16 +69,22 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
       return $delegate;
     });
   })
-  .run(function($http, jhConfig) {
+  .run(function($rootScope,$http, jhConfig) {
+    //global function to replace the placeholders
+    $rootScope.replacePlaceholder = function(value){
+      return value.replace("%{hostname}",window.location.host)
+    }
+    //function ends here
+
     var request = new XMLHttpRequest();
     request.open('GET', 'config.json', false);
     request.send(null);
     if (request.status === 200) {
       var config = JSON.parse(request.responseText);
       angular.forEach(config, function(value, key) {
-         jhConfig[key] = value.replace("%{hostname}",window.location.host);
+        //assigning config value with replaced value of placeholder
+         jhConfig[key] = $rootScope.replacePlaceholder(value);
       });
-      console.log(jhConfig);
     } else {
       console.warn('No configuration found');
     }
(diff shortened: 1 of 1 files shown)


src/app/index.js
```diff
@@ -76,8 +76,9 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
     if (request.status === 200) {
       var config = JSON.parse(request.responseText);
       angular.forEach(config, function(value, key) {
-         jhConfig[key] = value;
+         jhConfig[key] = value.replace("%{hostname}",window.location.host);
```
imobachgs: What about extracting this logic to a separate function? Just in case we want to add more placeholders.

src/app/index.js
```diff
@@ -76,8 +76,9 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
...
       });
+      console.log(jhConfig);
```
imobachgs: Is this one a left-over from a debugging session? If not, just please add some description so you know what this object is. Something like this:

```javascript
console.log("Processed configuration: ", jhConfig);
```

Sure you find a better name.

mohsini172: Ohhh yes its due to debugging will change that.

src/app/index.js
mohsini172: Ok I will separate that.

mohsini172:
please review now

src/app/index.js
```diff
@@ -69,16 +69,22 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
       return $delegate;
     });
   })
-  .run(function($http, jhConfig) {
+  .run(function($rootScope,$http, jhConfig) {
+    //global function to replace the placeholders
+    $rootScope.replacePlaceholder = function(value){
```
imobachgs: Why are you attaching this function to the `$rootScope`? What about just declaring and using the function? So you won't need to inject the `$rootScope` here.

src/app/index.js
```diff
@@ -69,16 +69,22 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
...
+      return value.replace("%{hostname}",window.location.host)
```
imobachgs: Please, observe style rules:

* Use a space to separate arguments: `%{hostname},window` to `%{hostname}, window`.
* Uses semicolon at the end of the line.

You can check the output of `gulp scripts` to find out style problems. On the other hand, when PR https://github.com/jangouts/jangouts/pull/178 is merged, you'll be able to run `gulp lint` to check your code style.

src/app/index.js
```diff
@@ -69,16 +69,22 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
...
+    }
+    //function ends here
```
imobachgs: It's a one line function, so I think this comment is not needed.

mohsini172: Ok got it

src/app/index.js
mohsini172: noted

src/app/index.js
mohsini172: I'm using the root scope to make it globally accessible but I will change that if you want.

imobachgs: My point is that you don't need it to be publicly accessible, so you don't need to pollute nor inject the `$rootScope`.

imobachgs:
@mohsini172 Thanks for your contribution!

mohsini172:
You're welcome. Glad to work :)

PR #180, opened 2017-03-10
Solves Typeerror

### LINKED ISSUE #179: Server not running on localhost due to 'Typeerror'
@imobachgs @ancorgs 
in src/app/index.js
```
    function replacePlaceholder(value) {
      return value.replace("%{hostname}", window.location.host);
    }
    var request = new XMLHttpRequest();
    request.open('GET', 'config.json', false);
    request.send(null);
    if (request.status === 200) {
      var config = JSON.parse(request.responseText);
      angular.forEach(config, function (value, key) {
        //assigning config value with replaced value of placeholder
            jhConfig[key] = replacePlaceholder(value);
      });
    } else {
      console.warn('No configuration found');
    }
```
The 'config' object may have non-string values. Therefore, if we pass such values to 'replacePlaceholder()' function, it will raise a "Typeerror" since 'value' is not a 'string', and, .replace() method works only for 'string' types.

code-master5:
Fixes #179 
Added a typecheck inside `replacePlaceholder()` function before returning the value.

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -71,9 +71,15 @@ angular.module('janusHangouts', ['ngAnimate', 'ngCookies', 'ngTouch',
   })
   .run(function($http, jhConfig) {
     //function to replace the placeholders
+    
     function replacePlaceholder(value) {
-      return value.replace("%{hostname}", window.location.host);
+      if (typeof value === 'string') {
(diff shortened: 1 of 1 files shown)


imobachgs:
Good point. Thanks!

PR #181, opened 2017-03-11
Has Left room was showing two times and has been fixed

### LINKED ISSUE #171: "Has left the room" message displayed twice
When the user OneUser quits, the other participants see the message "OneUser has left the room" twice. That should not be the case.

mohsini172:
#171 When a user leaves the room, the action service was calling the destroy feed two times one for leaving and one for unPublish.

### DIFF
--- a/src/app/components/room/action.service.js
+++ b/src/app/components/room/action.service.js
@@ -71,7 +71,7 @@
       LogService.add(entry);
     }
 
-    function destroyFeed(feedId) {
+    function destroyFeed(feedId,destroyType) {
       var feed = FeedsService.find(feedId);
       if (feed === null) { return; }
       $timeout(function () {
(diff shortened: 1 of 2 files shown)


mohsini172:
@imobachgs please review changes

ancorgs:
Not sure if adding a parameter to `ActionService.destroyFeed` is the right thing to do here. I would say we actually need two separate actions. One for unpublishing and one for leaving.

mohsini172:
Ok I will check the code if both of them can be separated.

imobachgs:
Thanks for your contribution @mohsini172.

Are you still interested in fixing this issue? Maybe you would like to cooperate with @aniabui in order to fix it. If you're not interested, just close the PR.

Thanks in advance!

mohsini172:
Yeah I will fix it within two days.

mohsini172:
@imobachgs Please review changes now.

src/app/components/room/action.service.js
```diff
@@ -83,6 +84,15 @@
       LogService.add(entry);
```
ancorgs: I would go one step further and re-implement `destroyFeed`to look like that:
```javascript
function destroyFeed(feedId) {
       unpublishFeed(feedId);
       // Log the event
       var entry = new LogEntry("destroyFeed", {feed: feed});
       LogService.add(entry);
}
```

That will reduce code duplication.

ancorgs: Oh, no. I was wrong. Never mind.

Issue #130, opened 2016-03-14
PIN code room


nlienard:
Hi

I'm using Janus as gateway, and it works perfectly with Jangouts (and thanks for that).

Now; i got a test room configured with a PIN:

[12341]
description = pin test
pin = 12345
publishers = 6
bitrate = 128000
fir_freq = 10
record = false

When trying to go to this room through Jangouts; there is a popup with following message : 

Missing element (pin)

Is there any way to implement the PIN code access into room ?

Thanks a lot !

code-master5:
@ancorgs I want to work on this. Can you please guide me a bit on this???

ancorgs:
Fixed by #182

PR #182, opened 2017-03-11
Pin code room Support Added

code-master5:
Enhancement for #130 .

### DIFF
--- a/src/app/components/room/room.service.js
+++ b/src/app/components/room/room.service.js
@@ -13,7 +13,7 @@
 
     RoomService.$inject = ['$q', '$rootScope', '$timeout', 'FeedsService', 'Room',
       'FeedConnection', 'DataChannelService', 'ActionService', 'jhConfig',
-      'ScreenShareService', 'RequestService'];
+      'ScreenShareService', 'RequestService', 'UserService'];
 
   /**
    * Service to communication with janus room
@@ -22,7 +22,7 @@
    */
   function RoomService($q, $rootScope, $timeout, FeedsService, Room,
       FeedConnection, DataChannelService, ActionService, jhConfig,
-      ScreenShareService, RequestService) {
+      ScreenShareService, RequestService, UserService) {
     this.enter = enter;
     this.leave = leave;
     this.setRoom = setRoom;
@@ -113,7 +113,7 @@
           // Step 1. Right after attaching to the plugin, we send a
           // request to join
           connection = new FeedConnection(pluginHandle, that.room.id, "main");
-          connection.register(username);
+          connection.register(username, UserService.getPin());
         },
         error: function(error) {
           console.error("Error attaching plugin... " + error);
@@ -213,7 +213,6 @@
           pluginHandle.send({"message": request, success: function(result) {
             // Free the resource (it looks safe to do it here)
             pluginHandle.detach();
-
             if (result.videoroom === "success") {
               var rooms = _.map(result.list, function(r) {
                 return new Room(r);
@@ -294,7 +293,7 @@
         plugin: "janus.plugin.videoroom",
         success: function(pluginHandle) {
           connection = new FeedConnection(pluginHandle, that.room.id, "subscriber");
-          connection.listen(id);
+          connection.listen(id, UserService.getPin());
         },
         error: function(error) {
           console.error("  -- Error attaching plugin... " + error);
@@ -359,7 +358,7 @@
         plugin: "janus.plugin.videoroom",
         success: function(pluginHandle) {
           connection = new FeedConnection(pluginHandle, that.room.id, videoSource);
-          connection.register(display);
+          connection.register(display, UserService.getPin());
           ScreenShareService.setInProgress(true);
         },
         error: function(error) {
--- a/src/app/components/feed/feed-connection.factory.js
+++ b/src/app/components/feed/feed-connection.factory.js
@@ -42,13 +42,13 @@
         this.pluginHandle.detach();
       };
 
-      this.register = function(display) {
-        var register = { "request": "join", "room": roomId, "ptype": "publisher", "display": display };
+      this.register = function(display, pin) {
+        var register = { "request": "join", "room": roomId, "ptype": "publisher", "display": display, "pin": pin || "" };
         pluginHandle.send({"message": register});
       };
 
-      this.listen = function(feedId) {
-        var listen = { "request": "join", "room": roomId, "ptype": "listener", "feed": feedId };
+      this.listen = function(feedId, pin) {
+        var listen = { "request": "join", "room": roomId, "ptype": "listener", "feed": feedId, "pin": pin || "" };
         pluginHandle.send({"message": listen});
       };
 
@@ -151,6 +151,7 @@
         } else {
           this.config = new ConnectionConfig(pluginHandle, options.values, null, options.ok);
         }
+        
       };
 
       /**
--- a/src/app/components/user/jh-signin-form.directive.js
+++ b/src/app/components/user/jh-signin-form.directive.js
@@ -47,6 +47,7 @@
       vm.room = null;
       vm.rooms = [];
       vm.listRooms = null;
+      vm.enteredPin = null; // pin entered by user
 
       RoomService.getRooms().then(function(rooms) {
         vm.room = RoomService.getRoom();
@@ -67,6 +68,7 @@
 
       function signin() {
         if (vm.room && vm.username){
+          UserService.setPin(vm.enteredPin);
           $state.go('room', {room: vm.room.id, user: vm.username});
         }
       }
--- a/src/app/components/user/jh-signin-form.html
+++ b/src/app/components/user/jh-signin-form.html
@@ -27,6 +27,17 @@
         <option value="" disabled selected>Select a room</option>
       </select>
     </div>
+    <div ng-show="vm.room.pin_required" class="form-group">
+      <label class="sr-only" for="inputPin">Pin</label>
+      <span>Please enter the right Pin to access this Room!</span>
+        <input
+          id="inputPin"
+          class="form-control input-lg"
+          type="password"
+          ng-model="vm.enteredPin"
+          placeholder="Pin (required)"
+          >
+    </div>
     <jh-thumbnails-mode-button text-template="true"></jh-thumbnails-mode-button>
     <button id="signInButton" class="btn btn-lg btn-primary btn-block" type="submit">Sign in</button>
   </form>
--- a/src/app/components/user/user.service.js
+++ b/src/app/components/user/user.service.js
@@ -93,5 +93,24 @@
     this.storeSettings = function() {
       localStorageService.set(USER_SETTINGS_KEY, this.settings);
     };
+    
+    // Pin entered by user
+    this.enteredPin = null;
+    /*
+     * Returns the pin as entered by the current user.
+     * @returns string A string representing the user like
+     *                
+     */
+    this.getPin = function() {
+      return this.enteredPin;
+    };
+
+    /*
+     * Set the pin to value entered by current user.
+     * @param   {string} val User entered pin.
+     */
+    this.setPin = function(value) {
+      this.enteredPin = value;
+    };
   }
 })();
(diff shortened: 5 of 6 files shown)


code-master5:
@imobachgs @ancorgs
This PR adds support for PIN rooms as follows:

* A new input box has been added in `src/app/components/user/jh-signin-form.html`. This will allow user to enter pin for room. This will ofcourse be optional as room might not require any pin for access.
* A `PinService()` has been added in `src/app/components/user/pin.service.js` (new file), that has methods for storing pin entered by user and getting the stored pin as required.
* The `src/app/components/user/jh-signin-form.directive.js ` depends upon `PinService()` for setting the pin entered by user.
* The `src/app/components/room/room.service.js ` uses `PinService()` for accessing the currently pin entered by the user. Then it sends this pin as an argument to methods inside ` src/app/components/feed/feed-connection.factory.js `.
* The  ` src/app/components/feed/feed-connection.factory.js ` has methods 
    1. `this.register()` - here `register` object requires `pin` property for room access.
    2.  `this.listen()` -  here `listen` object requires `pin` property for feed publish.

code-master5:
@imobachgs @ancorgs 
You can see the pin code room support in action here:
https://www.useloom.com/share/ed029ea3f3af465cac103544be4a7e1c

imobachgs:
@code-master5 Thanks for your contribution.

As a first attempt it looks ok, but we need to improve the UX before merging. I see two main issues:

1. Displaying the input control for the PIN code looks confusing because users could think that they need to specify one (although it says 'if required').
2. Loading the chat window and showing a pop-up when PIN is not correct is weird.

To solve 1) we could check if it's possible to know if a PIN is required in advance (when we fetch the list of rooms). Maybe we don't have that information, but we should check it.

In order to solve 2), we should not redirect the user to the videochat window until we have checked the PIN and connected successfully. If the PIN is wrong, we should notify the error in the login form (without pop-ups).

To be honest, I'm not sure about how hard it could be, so as a first step some research should be done.

Anyway, nice job :) Thanks!

src/app/components/room/room.service.js
```diff
@@ -213,7 +213,7 @@
           pluginHandle.send({"message": request, success: function(result) {
             // Free the resource (it looks safe to do it here)
             pluginHandle.detach();
-
+            console.log("This is the result", result);
```
imobachgs: If you think it's useful even in production, please improve the description. If not, just remove this line.

src/app/components/room/room.service.js
```diff
@@ -232,7 +232,7 @@
     // Enter the room
     function enter(username) {
       var deferred = $q.defer();
-
+
```
imobachgs: Clean up this spaces/tabs.

src/app/components/room/room.service.js
```diff
@@ -39,7 +39,7 @@
     this.room = null;
     this.rooms = null;
     this.janus = null;
-
+
```
imobachgs: Clean up this spaces/tabs.

src/app/components/user/pin.service.js
```diff
@@ -0,0 +1,36 @@
...
+ * of the MIT license.  See the LICENSE.txt file for details.
+ */
+
+(function () {
+  'use strict';
+
+  angular.module('janusHangouts')
+    .service('PinService', PinService);
+
+  function PinService() {
```
imobachgs: Why do you need a PinService at all? I think this information should belong to user.

code-master5:
@imobachgs Thanks for you review. Actually the UX you suggested is quite standard and was actually my first approach, but it I guess it required some serious changes  (like `we should not redirect the user to the videochat window until we have checked the PIN and connected successfully`).
As far as the check for `pin required or not` goes, I agree that we need to refer the Janus API documentation for whether there is a way to do that or not. To be quite frank, I could not figure out a method for this as the response object had no Pin field or anything like that. But I'll be digging into that. It'll be helpful if you provide some link for API documentation, specifically for the janus-gateway version used in `master branch`. I googled for documentation but I guess I'm getting one for other version.

src/app/components/user/pin.service.js
code-master5: @imobachgs Thanks. I'll merge the `UserService` and `PinService`.

imobachgs:
@code-master5 Maybe it's not supported in the current Jangouts API (I didn't checked myself). If we need to do some major change, we could document it and then start discussing how to handle it.

src/app/components/user/pin.service.js
code-master5: @imobachgs Actually, I chose a seperate service for `Pin` because `UserService` uses `angular-local-storage` module for storing user settings. `Pin` should not be stored, I guess. Although, I could store `Pin` locally in a variable, but then, IMHO It would be a bit out of flow. What you think??
Otherwise it's fine for me to merge both if you say so.

code-master5:
@imobachgs  I've been looking into Jangouts API docs here - https://www.linux-projects.org/documentation/rest-api/. It's clear that the object containing the list of rooms that we get does not contain any information regarding whether a particular room requires pin for entry or not. Infact we can only come to know this if we try to send the `join request object` with optional `pin` attribute inside `body` attribute. The server then sends a response, which may be `error` or `ack` depending on the `pin` attribute's value set in the request body.

code-master5:
@imobachgs Therefore, for the first task (`to know if a PIN is required in advance`), we need to have some workaround. My idea is:
     1. After getting `object containing the list of rooms`, we can send a `Dummy request object` to the server with `pin` set to null (and other attributes set to some random values). 
     2. Then if we get the `error` response, we can check the `type of error`. If it's the `pin` error, then we can set the `pinRequired` (or some suitable variable) to `true`. Depending upon the boolean value of `pinRequired`, we can show or hide the `Pin Input` box.

What are your thoughts on this??

code-master5:
@imobachgs But I also see some potential problems with this approach:
    1. For multiple rooms, we may have to send multiple `Dummy request objects`. It may take some time depending on the number of rooms available on the list. This may affect users' experience with Jangouts.
    2. We are assuming that a running Janus instance is atomic (in `room configuration` context) i.e. it reads the configuration once and stores it. Then it only reads room configuration once it's restarted. If this is not the case, then, if some admin resets the `pin` requirement for a room, the users may still be forced to enter `pin`, or, a room requiring `pin` may not show any input box for entering `pin`.
    3. Again, if we assume that admin may dynamically change room configurations while a Janus instance is running, then we will have to periodically check whether the room requires pin or not.

ancorgs:
Janus is also a free software project and the developers has been extremely helpful to Jangouts in the past (and they continue to be). So if we need an enhancement in the Janus API I would try to open an issue in the Janus repository so we discuss directly with the developers about the possibility to enrich their API. If they agree, they will probably add the enhancement themselves as time permits, and will allow us to open a pull request if we want to speed up things in our own benefit.

ancorgs:
Another option is not presenting the PIN box initially. Then, try to login without PIN. If it fails with the corresponding error, then present the PIN box to the user and retry.

Doing that, we don't need to know in advance if a room needs PIN. We will discover that at the right time and not before.

code-master5:
@ancorgs I think the approach given by you would be a bit awkward from users' point of view. Because first, the user will click on submit button and then user will get the input box asking for pin, in the worst case scenario. 
I think instead we should give the pin input box to the user initially and show that it's optional. Currently this is what the PR does. But for the second part where the user gets redirected even if the pin is incorrect, rather than an annoying popup, it would be better if we handle the `error` response and show message to the user that the entered pin is wrong (without redirecting to the room ofcourse).
What you say???

code-master5:
@imobachgs Ok then, I'll make the changes as suggested by @ancorgs , and, then send a PR. Thanks to both of you.

code-master5:
@imobachgs @ancorgs I've done the changes as you both suggested. Please review now. Changes are as follows:
  1. `UserService` and `PinService` have been merged into `UserService`.
  2. The `sign in` page does not have `pin input box` initially. The html code for `pin input box` has been added to `jh-signin-form.html`.
  3. `room.controller.js` - now instead of showing alert message in case pin required or wrong, it redirects the user to `sign in` page with `pin input box` visible along with an `error message` that the room requires pin for access or entry.

code-master5:
@imobachgs You can see the changes here:

https://www.useloom.com/share/e94d0d86431e4646aa42d5ee5e0bb409

ancorgs:
Now it's better, but I still find the interaction with the user a little bit weird. We probably need an intermediate state to make it look ok... which probably makes this a not-so-junior-job. 😉

code-master5:
@ancorgs I had an idea that you're gonna ask for this. I'll try to figure this one out. But I've also started working on my GSOC proposal (as @imobachgs suggested me to focus on that right now on IRC). So may be this might take some time. 
Thanks for your review. I was waiting for this one.
I just wanted one clarification on this. Is it like you want me to add another page on which the user should be redirected if the room she's trying to access requires pin??

code-master5:
@imobachgs @ancorgs May be I'm wrong but I feel that an intermediate state is not gonna help. This is because we cannot come to know of pin requirement for a room before trying to access the room.

Actually, the current scenario is that, when we try to access a room, it sends a response object. This object may be an error object or request granted object. In our case, we need to show the `pin required input box` only if it is required. For this, first we need to check whether room requires pin or not. Currently, the only way to do this is, after getting error response object, we can read the type of error. If the error type is `pin incorrect`, then only we show the `pin required input box`. 
Hence, even if we create an intermediate state for `pin requirement`, we need to enter the `room` state to recieve an error object and read the error message type. Hence the redirection will persist, as in this PR.

Please let me know if I'm not understanding what @ancorgs is trying to suggest.

Otherwise, I've also opened a [`Feature Request`](https://github.com/meetecho/janus-gateway/issues/826) at janus-gateway repository. This feature is gonna be of a lot of help for this issue.

ancorgs:
@code-master5 Opening that feature request for Janus definitely sounds to me like the way to go. Janus developers are really helpful folks. Well done. Why don't you try to go even one step closer and create a pull request for Janus implementing it? :wink:

code-master5:
@ancorgs I think I can try. They recently replied to add this to their TODO list and work on this feature. So, may be they would add this feature before I'll be able to understand the code-base. But there is nothing wrong in trying. In fact, I guess this will help me in getting a better understanding of the `janus-gateway` (as I have mentioned this in my GSOC 2017 Proposal).

code-master5:
@ancorgs @imobachgs `janus-gateway` folks are truly awesome. They added the functionality in just 2 days of opening Feature Request. So now I have added the required functionality. Now, the `pin requirement input box` will be shown as soon as user selects a `pin protected room` from rooms list. This looks awesome to me. There is no redirection required. If user enters a wrong pin, she will be redirected to sign in page. Please let me know if further changes are required.

code-master5:
@imobachgs @ancorgs 
To use `pin-required` featured, I had to build the latest jangouts. 

I wanted to ask if I need to update the `bower.json` to pull the latest `janus-gateway` repository? Actually, I tried this and jangouts did not work. Then I reverted the `bower.json` entry to fetch `janus-gateway` as earlier. This time jangouts worked! It's clear that the janus instance running is latest, and, the one being fetched by bower using `bower.json` is older. Can you explaing how is this working??

code-master5:
@ancorgs @imobachgs Please don't merge this. I'm facing some problems integrating the latest Janus-gateway with the master branch. More specifically, I'm not receiving chat messages. The video is working fine though.

code-master5:
@ancorgs @imobachgs Please merge this PR only after merging #188 . 👍

ancorgs:
@code-master5 in order to include this in the upcoming release it would be nice if you could rebase this on top of the current master, leaving out the (unneeded) changes to bower.json.

Or even better, open a new clean alternative pull request adding only the proper changes on top of master.

code-master5:
@ancorgs Please check out the changes now.

ancorgs:
I just tested this on top of the current master and it works as expected, as far as I can see.

So if @imobachgs' concerns about the code organization are cleared up, I will merge and include this in the upcoming release.

code-master5:
@imobachgs what are your views on this PR?? Please let me know if you feel the need for some changes.

PR #183, opened 2017-03-12
Username Change Support Added

code-master5:
Enhancement for #15


code-master5:
@imobachgs @ancorgs 
This PR adds Username Change Functionality as follows:

* A `src/app/components/feed/buttons/jh-nameedit-button.html` file has been added that defines UI for name change.
* A `src/app/components/feed/buttons/jh-nameedit-button.directive.js` file has been added. It defines the directive and controller. It uses `FeedsService` for getting the current publisher detail object. It also uses `$state` for changing the current publisher's `URL`.
* Following functions has been added to `src/app/components/feed/feeds.factory.js` -
    1. `updateDisplay()` - It takes the new display name, calls `setDisplay()` to set the new display name, and, calls `DataChannelService.sendStatus()` for notifying changes to remote peers.
    2. `getDisplay()` - It is used by `getStatus()`, which, in turn, is used by `DataChannelService` for reading current feed status.
    3. `setDisplay()` -  It sets the current display to new display as entered by the user.
* A `jh-nameedit-button` element has been added in `src/app/components/feed/jh-feed.html`. This button allows user to enter the edit mode.
* Finally, some styles for edit mode input has been added in `src/app/index.scss`.

code-master5:
@imobachgs @ancorgs 
This video shows the functionality in action:

https://www.useloom.com/share/a88becfb0ceb4c8998d4a28ab720e9d3

src/app/components/feed/buttons/jh-nameedit-button.html
```diff
@@ -0,0 +1,36 @@
...
+<button
+  type="button"
+  class="btn btn-default"
+  ng-click="vm.toggleEditingMode()"
+  title="Edit Name"
+  ng-show="vm.isEditable()">
+  <span class="glyphicon glyphicon-edit"></span>
+</button>
+
+<!--
```
imobachgs: Please, remove dead code.

code-master5:
@imobachgs Thanks for your review. Actually in the area where participants are visible, we have a toggle button for each participant which gives options like:
    1.  'Sticky' and 'disable video' for the publisher.
    2. 'Sticky', 'Ignore' and 'Mute' for others.
in that order.
As you can see, there is a difference in number of buttons for publisher and others. So, I saw a good space for `name change button` here for publisher, as, infact only publisher should be allowed to change her name. As you can see, it fits well in the UI. Also when I decided to design this way, it seemed to go well according to the code pattern.
But I know it's more of a `design related decision` kinda thing. 
I'll be waiting for your and @ancorgs 's final decision on the design.

imobachgs:
As a user, I wouldn't expect the option to be there but maybe clicking on the name. If you prefer, we could consider clicking the name that is below the participant. @ancorgs Opinions?

ancorgs:
I would also expect the option to be in the top bar, next to the name. It simply looks like the natural and intuitive place to me.

That being said, I don't have a very strong opinion. I find the UI proposed in this PR to also be acceptable. We have a separate GSoC project to rethink the UI anyways.

code-master5:
@ancorgs Should I add the name change option in the top bar only, or both the places ???

imobachgs:
IMHO one place is enough.

code-master5:
@imobachgs Which place ?? Top bar ??

imobachgs:
Yes, top bar.

code-master5:
@imobachgs Ok then. I'll make changes and send PR. Thanks to both of you for your valuable reviews and time.

src/app/components/feed/jh-main-feed-info.directive.js
```diff
@@ -27,12 +27,77 @@
...
+        if (event.keyCode === 13) {
+          vm.updateName();
+          vm.leaveEditingMode();
+        // if keypressed is 'Escape'
+        // then leave the editing mode without changing name
+        } else if (event.keyCode === 27) {
+          vm.leaveEditingMode();
+        }
+      }
+
```
code-master5: `keyDetect()` function listens for a keypress event in the input box. If `Enter` is pressed the name is changed to the current value in input box and it exits editing mode. If `Escape` is pressed, it only leaves editing mode and does not update name. 
Other functions are same as explained earlier.

src/app/components/feed/jh-main-feed-info.html
```diff
@@ -1 +1,15 @@
...
+    <span class="glyphicon glyphicon-question-sign"
+      title="Press Enter to change &#013; Press Escape to keep"></span>
+  </div>
+  <span class="feed-info"
+  title="Double Click to Edit Name"
+  ng-dblclick="vm.makeNameEditable()"
+  ng-hide="vm.isEditing()">
+  {{ vm.name() }}
+  </span>
+</span>
```
code-master5: A div `class="name-edit"` has been added here. It shows only when user double clicks on the name. There is an input field inside it that allows users to change name. The span inside this div is a `help` icon. When a user hover over this icon, it shows the instructions for changing name.

The `feed-info` span shows only when user is not editing.

code-master5:
@imobachgs Please review now. I've done the changes as you suggested. I've removed `src/app/components/feed/buttons/jh-nameedit-button.html` and `src/app/components/feed/buttons/jh-nameedit-button.directive.js` files.

code-master5:
@imobachgs You can see the changes in action here:
https://www.useloom.com/share/0475dd0ee41b42ac95df18156c553ef8

ancorgs:
In general I like it. Just a minor comment: would it be possible to set focus on the textbox right away when it appears. Having to click on the name to turn it editable and then again in the text field to start writing feels weird.

Other than that, I consider this good enough to be merged.

code-master5:
@ancorgs Yeah, you're right! I suddenly realised that this is actually a bit of frustating from user's point of view. I should've considered this in the first place.
All right then, I'm on it. Thanks for your review.

code-master5:
@ancorgs I've added the autofocus support. Surprisingly, there is no inbuilt way in AngularJS for conditional autofocusing on input box. So I had to add a new a auto-focus directive `auto-focus.directive.js` file.

ancorgs:
Now it works as I expected. This LGTM, so I would merge.

@imobachgs your review is still blocking the merge, so please take a look.

Issue #160, opened 2016-11-08
Main Video Fullscreen Controls


manno:
Currently one has to click on the video window and select "full screen" from the context menu of the HTML5 player. 

For one the button is kind of hard to reach. However, for screen sharing it would be great to keep the participant videos around, i.e. by scaling the main video to the pages background and layering the portrait videos on top.

code-master5:
@ancorgs I am working on this.

Issue #22, opened 2015-06-03
per user volume control


aschnell:
Since some users are way to loud and others very quiet it should be possible to control the volume per user.

thardeck:
This would also allow local mute for people in the same room like Mumble.

manno:
My colleagues keep asking for something like this. They're all sitting in the same room and can hear each other without the headset. Apparently it's confusing to hear the same voice twice, with a delay.

ancorgs:
Firefox allows a trick to deal with people in the same room. If you right-click in the video thumbnail you will see a "mute" option that will only mute your particular stream of the speaker (that is, not muting the participant in the jangouts meaning of muting). That makes the trick. I'm considering to make that option more obvious and available in chrome as well. But we need to find the time and to redesign the UI a littler bit.

ancorgs:
@manno With Jangouts 0.4.5 it's even easier to mute your colleagues in the same physical room. Just click on their name and then in "don't play audio".

shundhammer:
An alternative would be to automatically normalize the audio volume: Reduce the loud ones and increase the soft ones to a common average level.

PR #185, opened 2017-03-23
Full Screen Video

bimalkant-lauhny:
Enhancement for #160 .

### DIFF
--- a/src/app/components/feed/jh-main-feed.directive.js
+++ b/src/app/components/feed/jh-main-feed.directive.js
@@ -10,8 +10,10 @@
 
   angular.module('janusHangouts')
     .directive('jhMainFeed', jhMainFeedDirective);
-
-  function jhMainFeedDirective() {
+  
+    jhMainFeedDirective.$inject = ['$timeout'];
+    
+  function jhMainFeedDirective($timeout) {
     return {
       restrict: 'EA',
       templateUrl: 'app/components/feed/jh-main-feed.html',
@@ -32,6 +34,41 @@
           video.muted = true;
           Janus.attachMediaStream(video, newVal);
         }
+      });
+      
+      $timeout(function(){
+        var mainFeed = document.getElementById('main-feed');
+        mainFeed.addEventListener('dblclick', function() {
+          // if video is in fullscreen mode
+          if (
+            document.fullscreenElement ||
+            document.webkitFullscreenElement ||
+            document.mozFullScreenElement ||
+            document.msFullscreenElement
+          ) {
+            if (document.exitFullscreen) {
+              document.exitFullscreen();
+            } else if (document.webkitExitFullscreen) {
+              document.webkitExitFullscreen();
+            } else if (document.mozCancelFullScreen) {
+              document.mozCancelFullScreen();
+            } else if (document.msExitFullscreen) {
+              document.msExitFullscreen();
+            }
+            // if video is not in fullscreen mode
+          } else {
+            if (mainFeed.requestFullscreen) {
+              mainFeed.requestFullscreen();
+            } else if (mainFeed.webkitRequestFullscreen) {
+              mainFeed.webkitRequestFullscreen();
+            } else if (mainFeed.mozRequestFullScreen) {
+              mainFeed.mozRequestFullScreen();
+            } else if (mainFeed.msRequestFullscreen) {
+              mainFeed.msRequestFullscreen();
+            }
+          } 
+        });
+        
       });
     }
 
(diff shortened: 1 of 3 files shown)


code-master5:
@imobachgs @ancorgs 
This PR adds full screen video toggle mode on double click.
1. A double click event listener has been added in `src/app/components/feed/jh-main-feed.directive.js`. This listener toggles full screen video mode on double click.
2. A tooltip for double click has been added in `src/app/components/feed/jh-main-feed.html`.

code-master5:
@imobachgs Thanks for the review. I'll look into that.
Actually, I was looking at issue #22 . Since we also wanted to provide volume control, I was thinking maybe we should just allow the HTML5 video controls by default. But the problem is that, the video is
rotated 180 deg horizontal by `mirrored` class

code-master5:
Sorry @imobachgs , I tried fullscreen API but it didn't work. It seems like HTML5 API doesn't support any styles on fullscreen video tag.

dgdavid:
Hi @code-master5,

Did you try to wrap the video inside another element and apply the `fullscreen` mode and necessary CSS transformations to it? Reading the [Fulscreen API MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) it seems to me that this could be a useful approach to solve the issue, since talking about `Presentation differences` they say:

> On the other hand, if you're trying to emulate WebKit's behavior on Gecko, you need to place the element you want to present inside another element, which you'll make fullscreen instead, and use CSS rules to adjust the inner element to match the appearance you want.

It is just an idea, I didn't try it.

code-master5:
@dgdavid  Thank you very much. It's working now. I already had tried that. But I was calling the fullscreen API function with element selected using jQuery. Thanks to your comment, I tried again. This time I used simple javascript selector syntax. It worked like charm (atleast in Firefox and Chrome).

code-master5:
@imobachgs Please review changes now. 
As @dgdavid suggested, I used a div `#main-feed` to wrap video tag. Now, a double click on video tag opens `#main-feed` in fullscreen mode. I have used CSS fullscreen pseudo-class syntax to stretch video to fullscreen only when `#main-feed` is in fullscreen mode.

src/app/components/feed/jh-main-feed.directive.js
```diff
@@ -33,8 +35,44 @@
           Janus.attachMediaStream(video, newVal);
         }
       });
+      
+      $timeout(function(){
+        var video = $('video', element)[0];
+        var mainFeed = document.getElementById('main-feed');
+        video.addEventListener('dblclick', function() {
```
dgdavid: Why not bind/catch the event directly to `main-feed`? Due to bubbling principle, when you'll make double click on `video` element it will be run also on its ancestors unless the propagation be stopped.

code-master5: @dgdavid All right, I'll do that if you say so. I don't know what I was thinking while doing that. Thanks for review and explanation.

code-master5:
@dgdavid I changed the code to bind the `dblclick` event to `mainFeed` directly. This works fine.

src/app/components/feed/jh-main-feed.directive.js
dgdavid: No worries. Your code was not wrong, my comment was just an opinion / improvement.

src/app/components/feed/jh-main-feed.directive.js
```diff
@@ -33,6 +35,41 @@
           Janus.attachMediaStream(video, newVal);
         }
       });
+      
+      $timeout(function(){
+        var mainFeed = document.getElementById('main-feed');
+        mainFeed.addEventListener('dblclick', function() {
+          // if video is in fullscreen mode
+          if (
```
dgdavid: I don't know right now how to organise the `utilities` functions in Angular 2, but it makes sense, at least for me, to move this code to a separated file/module that you could import and use here. 

Then, you can simplify the directive writing something like 

```javascript
mainFeed.addEventListener('dblclick', function() { toggleFullScreen(mainFeed) })
```

Even more, Iooking for a module example to illustrate my comment I found, accidentally, the [screenfull](https://github.com/sindresorhus/screenfull.js) library, a cross-browser wrapper that you could use if the maintainers allow you to include a small dependency.

dgdavid: Hmmm, I notice that you're making the PR in the `master` branch. So, I must refer to AngularJS instead Angular 2 :sweat_smile::sweat_smile::sweat_smile:

code-master5:
@ancorgs Hey! I've updated this too! Check this one out..

PR #186, opened 2017-04-02
Fix "Has left the room" message displayed twice

aniabui:
I split `destroyFeed()` into two functions: `destroyFeedLeaving()` and `destroyFeedUnpublish()`. It solves the issue #171 in a way suggested by ancorgs (https://github.com/jangouts/jangouts/pull/181#issuecomment-287382181).

I assumed that `destroyFeedUnpublish()` doesn't need any extra logs.

Also, I didn't live test it because I couldn't get Jangouts to run properly on my computer. I did `vagrant up`, `vagrant ssh` to login and then `gulp serve` to start a Jangouts server. I opened Jangouts on different browsers (Chrome, Firefox, Opera Neon), but clients cannot detect each other except for the login screen where they can detect the number of participants in a conference room.

I am not too familiar with Angular, but should we also consider making `destroyFeed()` private?

I am using macOS, Firefox 52.0.2, Opera Neon 1.0.2531.0 and Chrome 56.0.2924.87 (I turned off all plug-ins as well).

### DIFF
--- a/src/app/components/room/action.service.js
+++ b/src/app/components/room/action.service.js
@@ -82,17 +82,19 @@
     }
 
     function destroyFeedLeaving(feedId) {
-      destroyfeed(feedID)
+      var feed = FeedsService.find(feedId);
+      destroyFeed(feedId);
       // Log the event
       var entry = new LogEntry("destroyFeedLeaving", {feed: feed});
       LogService.add(entry);
     }
 
     function destroyFeedUnpublish(feedId) {
-      destroyfeed(feedID)
+      var feed = FeedsService.find(feedId);
+      destroyFeed(feedId);
       // Log the event
       var entry = new LogEntry("destroyFeedUnpublish", {feed: feed});
-      if (entry.text !="") {
+      if (entry.text !== "") {
         LogService.add(entry);
       }
     }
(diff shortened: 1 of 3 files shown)


aniabui:
I am not sure if `destroyFeedUnpublish` works well because I couldn't test it. Also, I am not sure if it is logically correct right now.

src/app/components/room/action.service.js
```diff
@@ -78,11 +79,24 @@
         feed.disconnect();
         FeedsService.destroy(feedId);
       });
+    }
+
+    function destroyFeedLeaving(feedId) {
+      destroyfeed(feedID)
       // Log the event
-      var entry = new LogEntry("destroyFeed", {feed: feed});
+      var entry = new LogEntry("destroyFeedLeaving", {feed: feed});
```
imobachgs: `feed` is not defined.

src/app/components/room/action.service.js
```diff
@@ -78,11 +79,24 @@
...
       LogService.add(entry);
     }
 
+    function destroyFeedUnpublish(feedId) {
+      destroyfeed(feedID)
```
imobachgs: You should use `destroyFeed` instead of `destroyfeed`.

aniabui:
@imobachgs I managed to make it work! I was connecting to it using http://0.0.0.0:3000 and realized that Chromium required me to use http://localhost:3000 instead. Here is the link to documentation: https://www.chromium.org/Home/chromium-security/prefer-secure-origins-for-powerful-new-features.

ancorgs:
Closed in favor of #181 which implements the same in slightly cleaner way (in my opinion).

PR #188, opened 2017-04-16
Update Jangouts to run with latest Janus-gateway

code-master5:
This PR can be considered as a major step towards solving the #187 issue.

PR #193, opened 2017-06-06
Work with recent versions of janus.js and janus-gateway

ancorgs:
As long as the janus-gateway in the server does not include [this commit](https://github.com/meetecho/janus-gateway/commit/db66959e4e462a3185f8f64ebd2fb5e5d014f0b2) (that is, Janus gateway must be older than April 28th).

But that's a different topic (a Janus bug, IMHO).


ancorgs:
replaces #193

code-master5:
@ancorgs Congrats on this! 👍 Should I close #188 ???

PR #194, opened 2017-06-06
Add link to Google Groups

aniabui:
I added a link to Google Groups in 'Find us' section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,5 +112,6 @@ This software is released under the terms of the MIT License. See the
 
 ## Find us
 
-Jangouts developers can be usually found at
-[#jangouts IRC channel at Freenode](irc://irc.freenode.net/jangouts)
+Jangouts developers can be usually found at:
+ - [#jangouts IRC channel at Freenode](https://webchat.freenode.net/?channels=%23jangouts)
+ - [Google Groups](https://groups.google.com/forum/#!forum/jangouts)


README.md
```diff
@@ -112,5 +112,6 @@ This software is released under the terms of the MIT License. See the
 
 ## Find us
 
-Jangouts developers can be usually found at
-[#jangouts IRC channel at Freenode](irc://irc.freenode.net/jangouts)
+Jangouts developers can be usually found at:
+ - [#jangouts IRC channel at Freenode](https://webchat.freenode.net/?channels=%23jangouts)
+ - [Google Groups](https://groups.google.com/forum/#!forum/jangouts)
```
imobachgs: Just a minor comment: I would say 'Google Group' (singular) or 'Mailing List' as you prefer.

ancorgs: Well, if we are going to look that close, :wink: I would even state that IRC is good to reach the developers at a given point in time and the Google group is more targeted for general discussions within the whole community. However you want to word that.

imobachgs: That's maybe too much :) Just changing it to *singular* is enough for me :)

aniabui: I just want to point out that the it's officially called Google GroupS, not Google Group :)

imobachgs: Yes, but Jangouts group is only one :) Anyway, let's keep it as it is :)

Thanks!

PR #201, opened 2017-07-24
Change footer

aniabui:
This branch is based on the redesign branch from my fork and that's why the commits from the previous pull-request #200  are visible here.

screen shot 2017-07-24 at 4 00 12 pm

PR #202, opened 2017-07-24
Redesign Navigation Bars

aniabui:
This PR depends on #200 & #201, which are not merged yet.


src/app/components/videochat/jh-video-chat.html
```diff
@@ -132,7 +132,6 @@
 
 <script>
 $("#menu-toggle").click(function(e) {
-  e.preventDefault();
```
aniabui: I am not sure if removing this is the best thing to do, but I don't think it makes any difference.

src/app/components/videochat/jh-video-chat.html
```diff
@@ -1,65 +1,137 @@
...
+          </li>
+        </ul>
+      </div>
+    </div>
+  </div> -->
 </div>
+
+<script>
+$("#menu-toggle").click(function(e) {
+  $("#wrapper").toggleClass("toggled");
```
aniabui: I removed this line from the script in between lines 134 & 135.

    e.preventDefault();

I am not sure if removing this is the best thing to do, but I don't think it makes any difference.

src/app/components/videochat/jh-video-chat.html
```diff
@@ -1,65 +1,137 @@
...
+          <jh-main-feed-info></jh-main-feed-info>
+          <jh-leave-button></jh-leave-button>
         </div>
-      </li>
-    </ul>
+      </div>
+    </header>
   </div>
+
+  <!-- Page Content -->
```
imobachgs: Is all this code intentionally commented?

aniabui: I only commented it out because it was easier to work with without all that noise there. Would you like me to uncomment it?

imobachgs: No no, it is ok for me.

PR #203, opened 2017-07-24
Change footer

aniabui:
This branch is an uncluttered version of PR #201.

screen shot 2017-07-24 at 4 00 12 pm


src/app/components/footer/jh-footer.jade
```diff
@@ -1,9 +1,16 @@
...
+  .container-fluid
+    span.pull-left#text
+      | &copy;
+      = " "
+      a(href='http://www.suse.com/', target='_blank') SUSE 2015
+      = " | "
+      | version
+      = " "
+      = version
+      = " | "
```
imobachgs: What about something like this?

```jade
  .container-fluid                                                                                                                                                                                                
    span.pull-left#text                                                                                                                                                                                           
      | &copy;&nbsp;                                                                                                                                                                                                
      a(href='http://www.suse.com/', target='_blank') SUSE 2015                                                                                                                                                   
      | &nbsp;| version #{version} |&nbsp;                                                                                                                                                                        
      a(href='https://github.com/jangouts/jangouts', target='_blank') GitHub
```

IMHO, it looks more readable.

aniabui:
@imobachgs g is intended to cover the text underneath. If this is not the best solution, we can work on this later as it's not that urgent I think.

imobachgs:
LGTM now. Thanks!

PR #205, opened 2017-07-28
Modify Info Button

aniabui:
This PR depends on #202.

image

image


cyntss:
There is an issue in ``` src/app/components/videochat/jh-video-chat.directive.js```:

image

You don't have to fix it now, but don't forget to clean it up later

src/app/components/videochat/jh-video-chat.html
```diff
@@ -1,65 +1,137 @@
...
+        <button
+           class="btn btn-sidebar"
+           <span><img class="btn-img" src="images/buttons/buttons_window-share.svg"/></span>
+        </button>
+      </div>
+      <!-- Screen Share Button -->
+      <div class="sidebar-box">
+        <button
+           class="btn btn-sidebar"
+           <span><img class="btn-img" src="images/buttons/buttons_screen-share.svg"/></span>
```
cyntss: the ```<span>``` is not needed to hold the ``````, you can remove it from all the other buttons :)

cyntss:
That is all I could find. Once that is fixed we can merge.

One minor UX issue. It is hard to understand what the icons mean, and opening the cheatsheet it really not the best UX since it is not aligned to the icons, so it makes it hard to findt the icon you are trying to figure out.

My suggestion is that we use tooltips for the icons (keeping the cheatsheet of course), so when I want to know what an icon means, as a user I would put the mouse over and it would give me the name of the icon (the whole explanation wouldnt be enough for this).
For example:

1. I put the mouse over the icon:
image

2. then I see the tooltip with the name:
image

aniabui:
@cyntss 

> That is all I could find. Once that is fixed we can merge.
> 
> One minor UX issue. It is hard to understand what the icons mean, and opening the cheatsheet it really not the best UX since it is not aligned to the icons, so it makes it hard to findt the icon you are trying to figure out.
> 
> My suggestion is that we use tooltips for the icons (keeping the cheatsheet of course), so when I want to know what an icon means, as a user I would put the mouse over and it would give me the name of the icon (the whole explanation wouldnt be enough for this).
> For example:
> 
> I put the mouse over the icon:
> image

> 
> then I see the tooltip with the name:
> image

Did you mean to match the tooltips with the descriptions in cheatsheet so they are more consistent? I see one problem with this. For example, when the user hovers over the Audio icon, should the tooltip be "audio on" (the current state) or "audio off" (the state after clicking on the button). I think that would be a little bit misleading as the user wouldn't know exactly if the audio is on or off at the moment. Currently, Jangouts has the commands/instruction like "disable video" or "enable video" and I think this solution is clearer for the user. We can perhaps even make it clearer and rephrase to "push to disable video," but I'm not sure if adding "push" would be redundant. What do you think?

cyntss:
Perfect! it is ready to be merged :)

PR #206, opened 2017-08-04
Redesign Chatbox

aniabui:
This PR is based on #205.

image

image

image

### DIFF
--- a/src/app/components/chat/jh-chat.directive.js
+++ b/src/app/components/chat/jh-chat.directive.js
@@ -44,7 +44,7 @@
           messagesList.scrollTop = messagesList.scrollHeight;
         }, 100);
 
-        // Update the chat message count
+        // Update the chat messages count
         for (var i = scope.entriesCount; i < entries.length; i++) {
           if (entries[i].type === 'chatMsg') {
             scope.messagesCount++;
@@ -52,12 +52,13 @@
               scope.lastSeenMessage++;
             } else {
               var chatHeader = document.getElementById('chat-header');
-              chatHeader.innerHTML = scope.messagesCount-scope.lastSeenMessage;
+              chatHeader.innerHTML = scope.messagesCount - scope.lastSeenMessage;
             }
           } else {
             // We don't notify the user when the entry is not of type 'chatMsg'
           }
         }
+        scope.entriesCount = entries.length;
       });
 
       scope.$watch('vm.isChatVisible', function(isVisible) {
@@ -73,7 +74,7 @@
           scope.isChatVisible = false;
           if (scope.messagesCount === 0) {
             chatHeader.innerHTML = "Join the chat!";
-          } else if (scope.messagesCount === 1) {
+          } else if ((scope.messagesCount - scope.lastSeenMessage) === 1) {
             chatHeader.innerHTML = "1 unread message";
           } else {
             chatHeader.innerHTML = (scope.messagesCount - scope.lastSeenMessage) + " unread messages";
--- a/src/app/components/chat/jh-chat.directive.js
+++ b/src/app/components/chat/jh-chat.directive.js
@@ -32,36 +32,37 @@
       scope.lastSeenMessage = 0;    // last seen entry of type 'chatMsg'
       scope.isChatVisible = false;
 
+      // Scroll to the bottom of messages list.
       scope.$watch('vm.logEntries().length', function() {
-        // Scroll to bottom of messages list.
         var messagesList = document.getElementById('chat-messages-box');
         setTimeout(function() {
           messagesList.scrollTop = messagesList.scrollHeight;
         }, 100);
       });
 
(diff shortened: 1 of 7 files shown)


src/app/components/chat/jh-chat.directive.js
```diff
@@ -25,18 +27,95 @@
...
+        for (var i = scope.entriesCount; i < entries.length; i++) {
+          if (entries[i].type === 'chatMsg') {
+            scope.messagesCount++;
+            if (scope.isChatVisible) {
+              scope.lastSeenMessage++;
+            } else {
+              var chatHeader = document.getElementById('chat-header');
+              chatHeader.innerHTML = scope.messagesCount - scope.lastSeenMessage;
+            }
+          } else {
```
imobachgs: Why not just remove this else? You can put a similar comment on the `if` if you do not want to miss this information.

src/app/components/chat/jh-chat.directive.js
```diff
@@ -25,18 +27,95 @@
...
+      });
+
+      scope.$watch('vm.logEntries()', function(entries) {
+        // True only at the beginning when there are no chat messages
+        if (entries.length === 0) {
+          return;
+        }
+
+        // Update the chat messages count
+        for (var i = scope.entriesCount; i < entries.length; i++) {
```
imobachgs: This code looks quite "procedural" and error prone. What about a more functional style? For instance, you could filter the messages you are interested (with `Array.prototype.filter`) and then use `length` to set counters.

src/app/components/chat/jh-chat.directive.js
```diff
@@ -25,18 +27,95 @@
...
+          if (unreadMessages === 0) {
+            $("#chat-header").toggleClass("read");
+          } else {
+            $("#chat-header").toggleClass("unread");
+          }
+        } else {
+          scope.isChatVisible = false;
+          if (unreadMessages === 0) {
+            chatHeader.innerHTML = "Join the chat!";
+            $("#chat-header").toggleClass("read");
```
imobachgs: Basically you are calling `toggleClass` in several places. What about setting some `cssClass` variable and just call `toggleClass` at the end? IMHO it would be a cleaner approach.

src/app/components/chat/jh-chat.directive.js
```diff
@@ -25,18 +27,95 @@
...
+          chatHeader.innerHTML = ""; // Display nothing when chat is open
+          if (unreadMessages === 0) {
+            $("#chat-header").toggleClass("read");
+          } else {
+            $("#chat-header").toggleClass("unread");
+          }
+        } else {
+          scope.isChatVisible = false;
+          if (unreadMessages === 0) {
+            chatHeader.innerHTML = "Join the chat!";
```
imobachgs: What about `Click to open the chat`? After all, I am already in the chat.

src/app/components/chat/jh-chat.directive.js
```diff
@@ -25,18 +27,95 @@
...
+            $("#chat-header").toggleClass("read");
+          } else if (unreadMessages === 1) {
+            chatHeader.innerHTML = "1 unread message";
```
imobachgs: The same for the content. You could just set a `content` variable and set `innerHTML` at the end (and only once).

src/app/components/chat/jh-chat.directive.js
```diff
@@ -47,19 +47,18 @@
...
-            if (scope.isChatVisible) {
-              scope.lastSeenMessage++;
-            } else {
-              var chatHeader = document.getElementById('chat-header');
-              chatHeader.innerHTML = scope.messagesCount - scope.lastSeenMessage;
-            }
+        entries.slice(scope.entriesCount).filter(function(entry) {
+          // We don't notify the user when the entry is not of type 'chatMsg'
+          return entry.type === "chatMsg";
+        }).forEach(function(entry) {
```
imobachgs: You could do something like:

```javascript
newEntries = entries.slice(...)
scope.messagesCount += newEntries.length
if (scope.isChatVisible) {
  scope.lastSeenMessage += newEntries.length
} else {
  # Updating chatHeader.
}
```

imobachgs: About chat header, instead of using `getElementById` and updating it manually, maybe you should use a variable in the scope (`scope.chatHeader`) and use it in the template.

Thinking about it, maybe you could use the same approach when updating CSS classes and all that stuff.

I mean: try to set proper values on the directive's scope and let the template to show the correct values.

aniabui:
@imobachgs there is an issue with the chat now, because it's blinking when the page is newly open. Do you know how to fix this problem?

imobachgs:
@aniabui I was not able to reproduce the problem. After your latest changes, I think we can merge this PR.

Thanks!

PR #207, opened 2017-08-18
Main Video and Thumbnails Layout

aniabui:
This PR is based on #206.

@cyntss could you take a look at the layout and let me know what you think about it? There is a small detail that bothers me; the white space (see in the screenshot below) that doesn't look too beautiful. I am not sure if this that important, but I would appreciate if you could give me some tips on how to improve this layout :)

screen shot 2017-08-18 at 12 38 30 pm

Update: I found a way to deal with this issue. This is how it looks now:

image

PR #208, opened 2017-08-21
Redesign Thumbnails

aniabui:
This PR is based on #207.


aniabui:
@cyntss what are your thoughts on implementing the "more options" for each thumbnail. At the moment, when we click on the thumbnail, it is pinned to the top and when we click on the carat, the options appear. My question is, should we keep it as it is or should we get rid of the option of pinning to the top by 1-click and let all the options appear once the user clicks on the thumbnail (like in the picture below)? Or do you have any other ideas on how to deal with this problem?

image

cyntss:
@aniabui I agree with you and think that leaving it would be the best option for the time being. The icons look very good btw!

PR #209, opened 2017-08-22
Added Event Emitter and Callstats Service

code-master5:
This PR adds:

    1.  Event emitter support to Jangouts 
I've used RxJS (thanks to @imobachgs for suggestion). The `events.provider.js` is a module that 
provides an interface in form of a `Subject` to which any external service can subscribe to receive 
events. These events, then, can be used by services according to their use cases.

    2.  Callstats Service Integration
I've added `callstats.provider.js` as a module that reads callstats user configuration from 
`callstats.config.json` file. It subscribes to the event emitter and recieves events, which, it then
sends to the callstats.io service.

### DIFF
--- a/src/app/callstats/callstats.config.json
+++ b/src/app/callstats/callstats.config.json
@@ -1,4 +1,0 @@
-{
-  "AppID": "724446896",
-  "AppSecret": "/BKw/++t+DXe:UCxvwuXdkE6AIuxHXRj5PnaKLNeAgNZdp6rGShdWNmc="
-}
--- a/src/app/callstats/callstats.config.json
+++ b/src/app/callstats/callstats.config.json
@@ -1,0 +1,4 @@
+{
+  "AppID": "724446896",
+  "AppSecret": "/BKw/++t+DXe:UCxvwuXdkE6AIuxHXRj5PnaKLNeAgNZdp6rGShdWNmc=" 
+}
--- a/src/app/callstats/callstats.config.json
+++ b/src/app/callstats/callstats.config.json
@@ -1,0 +1,4 @@
+{
+  "AppID": "724446896",
+  "AppSecret": "/BKw/++t+DXe:UCxvwuXdkE6AIuxHXRj5PnaKLNeAgNZdp6rGShdWNmc="
+}
--- a/src/app/components/feed/feeds.factory.js
+++ b/src/app/components/feed/feeds.factory.js
@@ -15,7 +15,7 @@
   angular.module('janusHangouts')
     .factory('Feed', feedFactory);
 
-  feedFactory.$inject = ['$timeout', 'DataChannelService', 'SpeakObserver', 
+  feedFactory.$inject = ['$timeout', 'DataChannelService', 'SpeakObserver',
                          'jhEventsProvider'];
 
   /**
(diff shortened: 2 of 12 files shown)


code-master5:
@ancorgs Please review this.

src/app/components/feed/feeds.factory.js
```diff
@@ -290,6 +292,26 @@
                 if (options.after) { options.after(); }
                 // Send the new status to remote peers
                 DataChannelService.sendStatus(that, {exclude: "picture"});
+
```
ancorgs: I would make this event more generic. Something like (just a possible example)
```
type: "channel",
data {
  channel: type,
  status: enabled,
  peerconnection: that.connection.pluginHandle.webrtcStuff.pc
}
```

The fact that video and audio are considered different (for example "paused" vs "muted") is very specific to callstats.io, not a real general difference for Jangouts. So that "if" should be in the callstats provider.

src/app/components/room/room.service.js
```diff
@@ -353,10 +428,27 @@
       var display = FeedsService.findMain().display;
       var connection;
       var id;
-
+      
+      // emit `screenshare` event 
+      jhEventsProvider.emitEvent({
+        type: "screenshare",
```
ancorgs: This event also looks very influenced by callstats API. I don't think we need a different event for publishing screensharing and for publishing camera, just a different argument to the same type of event. From the jangouts point of view both are basically the same event even if callstats has a dedicated screenshare start/stop events.

src/app/components/room/room.service.js
```diff
@@ -106,10 +106,33 @@
       var $$rootScope = $rootScope;
       var connection = null;
 
+      // adding username to jhEventsProvider
+      jhEventsProvider.username = username;
+      // adding room details to jhEventsProvider
+      jhEventsProvider.roomDesc = that.room.description;
```
ancorgs: I wouldn't initialize this here. I would make the callstats provider ask directly the information it needs to `UserService` and `RoomService`. This is another unnecessary coupling of our event emitter with the callstats take on things.

src/app/callstats/callstats.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "AppID": "724446896",
+  "AppSecret": "/BKw/++t+DXe:UCxvwuXdkE6AIuxHXRj5PnaKLNeAgNZdp6rGShdWNmc="
+}
```
ancorgs: This should not be part of the PR, of course.

src/app/components/feed/feeds.factory.js
code-master5: @ancorgs Done! Please check this.

src/app/components/room/room.service.js
ancorgs: @code-master5 you told me about this being not possible. But I did some experiments with the logic of `jsEventsProvider` and `Callstats` being placed in new components within the `janusHangouts` module (instead of defining their own modules) and I don't see why shouldn't it work.

code-master5: @ancorgs Sorry, I can't understand. Can you provide me the link to your experiments for reference? Things were not working in my case! May be I missed something here.

src/app/callstats/callstats.config.json
code-master5: @ancorgs Done!

code-master5:
@ancorgs Hey there! nice to hear from you, after so long. Finally, this is merged and I'm happy. I'm very thankful to you and @imobachgs for constant help and mentoring. It was fun working with you guys.

PR #231, opened 2019-06-26
Add connection config, events and log entries services to Janus api

imobachgs:


src/janus-api/models/connection-config.js
```diff
@@ -0,0 +1,118 @@
...
+        okCallback = options.ok;
+      },
+      error: function() {
+        requested = null;
+        console.error("Config request not sent");
+      }
+    });
+  }
+
+  return that;
```
dgdavid: Thinking a little bit about this _`that` pattern_, maybe we could - in a future - to avoid it simply returning the object at the end. I mean, 

```javascript
export const createConnectionConfig = function(pluginHandle, wantedInit, jsep, ok) {
  var current = {};
  var requested = null;
  var wanted = {audio: true, video: true, data: true};
  var okCallback = null;
  
  ...

  return {
    get: function() {
      return current;
    },

    confirm: function() {
      ...
    },
  }
}

PR #235, opened 2019-06-27
Add feeds service to Janus API

imobachgs:
This PR adds feeds related factories and services. The tests are rather incomplete, but the code should be a good starting point.

### DIFF
--- a/src/janus-api/feeds-service.js
+++ b/src/janus-api/feeds-service.js
@@ -1,0 +1,114 @@
+/**
+* Copyright (c) [2019] SUSE Linux
+*
+* This software may be modified and distributed under the terms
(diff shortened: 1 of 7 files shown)


src/janus-api/feeds-service.js
```diff
@@ -0,0 +1,114 @@
...
+* of the MIT license.  See the LICENSE.txt file for details.
+*/
+
+export const createFeedsService = () => {
+  let mainFeed = null;
+  let feeds = {};
+  let that = {};
+
+  /**
+   * @returns{Feed} gets feed with given id or null if not found
```
ancorgs: Missing space. Isn't it?

PR #238, opened 2019-06-27
Initialize state

dgdavid:
A basic initialization of the app state, which will be handle by [Redux]() state container. It could be considered as a simple starting point, which sketch

* room
* participants
* messages

Please, keep in mind that, after a brief discussion, we decided to structure the _state ecosystem_ following an _approximation_ to the [ducks modular approach proposal](https://github.com/erikras/ducks-modular-redux). 

  > ### Rules
  > 
  > A module...
  > 
  > 1. MUST `export default` a function called `reducer()`
  > 2. MUST `export` its action creators as functions
  > 3. MUST have action types in the form `npm-module-or-app/reducer/ACTION_TYPE`
  > 4. MAY export its action types as `UPPER_SNAKE_CASE`, if an external reducer needs to listen for them, or if it is a published reusable library.
  >
  >  https://github.com/erikras/ducks-modular-redux#rules

There are other approaches on top of that, using directories, but we want to start with a file per _duck_.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10605,6 +10605,29 @@
       "resolved": "https://registry.npmjs.org/react-is/-/react-is-16.8.6.tgz",
       "integrity": "sha512-aUk3bHfZ2bRSVFFbbeVS4i+lNPZr3/WM5jT2J5omUVV1zzcs1nAaf3l51ctA5FFvCRbhrH0bdAsRRQddFJZPtA=="
     },
+    "react-redux": {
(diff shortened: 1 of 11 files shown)


dgdavid:
A long article about React & Redux & Ducks (not completely read yet, but seems interesting)

https://medium.com/swlh/the-good-the-bad-of-react-redux-and-why-ducks-might-be-the-solution-1567d5bdc698

PR #243, opened 2019-06-27
Fix several errors in the feeds factory

imobachgs:
Fix several issues in the feeds factory. However, prettier introduced a lot of style changes, which makes this PR harder to review.

### DIFF
--- a/src/janus-api/models/feed.js
+++ b/src/janus-api/models/feed.js
@@ -42,7 +42,7 @@ export const createFeedFactory = (
    * Checks if a given channel is enabled
    *
    * @param {string} channel - "audio" or "video"
-   * @returns {boolean}
+   * @returns {boolean,null}
    */
   that.isEnabled = function(channel) {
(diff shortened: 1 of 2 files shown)


src/janus-api/models/feed.js
```diff
@@ -1,455 +1,471 @@
...
-    /**
-     * Disconnects from janus
-     */
-    that.disconnect = function() {
-      if (that.connection) {
-        that.connection.destroy();
-      }
-      if (speakObserver) {
-        speakObserver.destroy();
+        return null;
```
dgdavid: Documentation says that a boolean is returned. `return false`?

dgdavid:
> which makes this PR harder to review.

Indeed!

src/janus-api/models/feed.js
imobachgs: Good point. I prefer to keep the original behavior, so I am updating the docs.

PR #248, opened 2019-06-28
Emit events on muting

ancorgs:
Adding the corresponding `emitEvent` for both kinds of mutes (triggered by the user and triggered by a peer request).

Note: prettier added A LOT of noise.


ancorgs:
closed in favor of #249

PR #252, opened 2019-06-28
Feature/participants

imobachgs:


src/state/ducks/participants.js
```diff
@@ -48,6 +55,15 @@ const reducer = function(state = initialState, action) {
           return obj;
         }, {});
     }
+
+    case PARTICIPANT_STREAM_SET: {
+      const participant = {...payload, stream_timestamp: new Date()};
```
dgdavid: :thinking: 

I think that here `payload` it is just the `participantId`, isn't it?

Look at https://github.com/jangouts/jangouts/pull/252/files#diff-79eaa44ae06f9d2e1c13f535ff418660R21

If so, we need to look for the participant first, right?

imobachgs: Yes, you are right. Fixed.

Before merging, I would like to have a look to all tests, because some of them are broken now.

dgdavid: > Before merging, I would like to have a look to all tests

I also would like to reword my "WIP" commit :)

Issue #254, opened 2019-07-03
Components tests not working


imobachgs:
Components tests are not working because of this error: `Invariant Violation: could not find react-redux context value; please ensure the component is wrapped in a <Provider>`. To reproduce the problem, just run `npm test`.

Once this problem is fixed, we should enable them in `.travis.yml`.

imobachgs:
Fixed in #255.

PR #255, opened 2019-07-03
Fix tests

imobachgs:
Use Travis to run tests. However, components testing is disabled because of #254.

### DIFF
--- a/src/janus-api/__mocks__/index.js
+++ b/src/janus-api/__mocks__/index.js
@@ -1,0 +1,12 @@
+/**
+* Copyright (c) [2019] SUSE Linux
+*
+* This software may be modified and distributed under the terms
+* of the MIT license.  See the LICENSE.txt file for details.
+*/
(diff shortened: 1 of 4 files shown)


src/janus-api/__mocks__/index.js
```diff
@@ -0,0 +1,12 @@
+/**
+* Copyright (c) [2019] SUSE Linux
```
dgdavid: NP: probably Prettier will "complain" here because a missing space after the `*`. The same in the other mock.

dgdavid: Another note: the rest of files has a `2015-2019` in the copyright header.

imobachgs: Actually, this is a new file, so it is ok as it is IMHO. Different thing are those files under `janus-api`, although I know that we have modified other new files aswell.

PR #258, opened 2020-02-06
Fix components tests

imobachgs:
This PR fixes component tests. Bear in mind that they are rather simple and incomplete, but they are a good starting point.

* Fix components and data-service-channel tests.
* Use [react-testing-library](https://github.com/testing-library/react-testing-library) for testing.
* Update react, react-dom and react-test-renderer.
* Enable all tests in Travis.

### DIFF
--- a/src/components/App/App.test.js
+++ b/src/components/App/App.test.js
@@ -8,9 +8,15 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import { renderWithRedux } from '../../setupTests';
+import { act, screen } from '@testing-library/react';
 
-it('renders without crashing', () => {
-  const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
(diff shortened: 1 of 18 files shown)


src/components/App/App.test.js
```diff
@@ -8,9 +8,15 @@
...
 import ReactDOM from 'react-dom';
 import App from './App';
+import { renderWithRedux } from '../../setupTests';
+import { act, screen } from '@testing-library/react';
 
-it('renders without crashing', () => {
-  const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
-  ReactDOM.unmountComponentAtNode(div);
+jest.mock('../../janus-api');
```
dgdavid: I was going to use the [`automock`](https://jestjs.io/docs/en/configuration.html#automock-boolean) option instead of mocking it explicitly per test file, but unfortunately CRA does not allow it without ejecting.

Issue #256, opened 2019-07-05
Get the room to connect to from the URL


imobachgs:
See #259.

PR #259, opened 2020-02-07
[React] Login from URL

imobachgs:
Implements support for #256. At this point in time, there is no support for getting the name from the local storage.


imobachgs:
> That's nice not only for production but also to ease the development tada

Yes, you are right. That was the main reason to implement this feature now :)

Issue #262, opened 2020-02-08
Improve room layout (including chat and participants list)


imobachgs:
Work in the Room, Chat and Participants components to display the information in a proper way.

dgdavid:
At some point back in the time, we concluded that it would be better to have a handful of prefixed (good) layouts instead the current one based on gridster.js

To achieve that, I think that Flexbox and/or CSS Grid are our allies, but I don't master them (yet :stuck_out_tongue_winking_eye:). That's why during this [Hack Week 19](https://hackweek.suse.com/19/projects/) I invested some time playing with them with one thing in mind: keep the whole interface within the browser viewport size, using scroll only in the chat and participants<sup>*</sup> _panels_ (when needed). And, although I couldn't finish something polished enough, I like the idea more than before :)

With enough knowledge/practice with those two things, it should be not difficult

* To find an easy way to define (and use) layouts
* To implement a layout switcher or view mode
* To use the `resize` CSS property to allow resize those predefined _panels_ (I found it a little bit problematic because a CSS grid panel seems to shrink properly but not to grow beyond the initial amount of space)
* To use the localStorage to save those changes

So, please, think in the current layout as a PoC :) Hopefully, I'm going to keep polishing it in dribs and drabs from now on.

---
<sup>*</sup> For sure, the [current behavior of the participants list, shrinking the participant miniatures to avoid scroll](https://github.com/jangouts/jangouts/blob/28e213278b2d515d43f8c1173e3183f4ddebcf01/src/app/components/videochat/jh-video-chat.directive.js#L250), should be brought back. But, IMHO, it should be an option ;)

PR #269, opened 2020-02-11
[React] Enhance chat

dgdavid:
Related to #262

### DIFF
--- a/src/components/Chat/MessageForm/MessageForm.js
+++ b/src/components/Chat/MessageForm/MessageForm.js
@@ -23,8 +23,12 @@ function MessageForm() {
   const dispatch = useDispatch();
 
   return (
-    <form className="MessageForm" onSubmit={(event) => handleSubmit(event, dispatch, textInput)}>
-      <input type="text" id="text" ref={textInput} />
+    <form
+      data-testid="chatbox"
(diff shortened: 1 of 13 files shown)


src/janus-api/models/log-entry.js
```diff
@@ -21,10 +21,10 @@ export const createLogEntry = (type, content) => {
   that.muteRequestText = function() {
     var res;
 
-    if (that.content.source.isPublisher) {
```
dgdavid: @imobachgs We should review this together, please.

I'm not sure if that change is really necessary.

src/components/Chat/MessageForm/MessageForm.js
```diff
@@ -23,8 +23,12 @@ function MessageForm() {
   const dispatch = useDispatch();
 
   return (
-    <form className="MessageForm" onSubmit={(event) => handleSubmit(event, dispatch, textInput)}>
-      <input type="text" id="text" ref={textInput} />
+    <form
+      data-testid="chatbox"
```
dgdavid: I'd like to talk about test, [`getByTestId`](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid), and this `data-testid` attribute too, please.

PR #270, opened 2020-02-11
Add a (un)mute button

imobachgs:
Add an (un)mute button. The current button should be replaced with a nicer one.

### DIFF
--- a/src/components/MuteButton/MuteButton.css
+++ b/src/components/MuteButton/MuteButton.css
@@ -1,0 +1,6 @@
+/**
+ * Copyright (c) [2015-2020] SUSE Linux
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE.txt file for details.
+ */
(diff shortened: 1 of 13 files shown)


src/components/MuteButton/MuteButton.css
```diff
@@ -0,0 +1,6 @@
+/**
+ * Copyright (c) [2015-2020] SUSE Linux
```
dgdavid: NP: are we going to keep the 2015-present Copyright for new things?

src/components/MuteButton/MuteButton.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) [2020] SUSE Linux
```
dgdavid: NP: here we had only 2020

src/components/MuteButton/MuteButton.js
```diff
@@ -0,0 +1,25 @@
+/**
```
dgdavid: NOTE: I had a similar approach but placing the button in _src/components/Participants/Actions/Mute.js_ But that kind of (re)organization can be done later. In fact, I prefer to do it later.

dgdavid:
I left some notes, but let's go ahead and merge it, please.

PR #274, opened 2020-02-11
[React] Fix audio/video initial state

imobachgs:
Do not set initial value of audio/video in the redux's store.

### DIFF
--- a/src/state/ducks/participants.js
+++ b/src/state/ducks/participants.js
@@ -64,7 +64,7 @@ const reducer = function(state = initialState, action) {
   switch (type) {
     case PARTICIPANT_JOINED: {
       // TODO: check initial values for audio/video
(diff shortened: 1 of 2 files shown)


src/state/ducks/participants.js
```diff
@@ -64,7 +64,7 @@ const reducer = function(state = initialState, action) {
   switch (type) {
     case PARTICIPANT_JOINED: {
       // TODO: check initial values for audio/video
-      const participant = { ...payload, audio: true, video: true, stream_timestamp: null };
```
dgdavid: So far, we need `audio` attribute to set set the `Mute` button. Which is the plan?

imobachgs: the plan is that at that point of time, that value is not known, so I am not setting any value.

dgdavid: So, fine with me. And for Travis now ;)

PR #280, opened 2020-02-12
[React] Logout action

### LINKED ISSUE #263: Add logout action
Add a logout action which takes the user to the login screen again.

dgdavid:
Implements the logout action.

* Fixes #263 
* Create and exports explicitly the `history` object, necessary to perform the redirection after exiting from a room
* Use `Router` instead `BrowserRouter`
* :warning: Tests pending

### DIFF
--- a/src/components/App/App.js
+++ b/src/components/App/App.js
@@ -7,11 +7,12 @@
 
 import React from 'react';
 import { Provider } from 'react-redux';
-import { BrowserRouter, Route } from 'react-router-dom';
+import { Router, Route } from 'react-router-dom';
 
 import store from '../../state/store';
 import janusApi from '../../janus-api';
 import { addEventsHandlers } from './events-handler';
+import history from '../../utils/history';
 
 import Login from '../Login';
 import Room from '../Room';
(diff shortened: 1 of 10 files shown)


src/components/App/App.js
```diff
@@ -7,11 +7,12 @@
 
 import React from 'react';
 import { Provider } from 'react-redux';
-import { BrowserRouter, Route } from 'react-router-dom';
+import { Router, Route } from 'react-router-dom';
 
 import store from '../../state/store';
 import janusApi from '../../janus-api';
 import { addEventsHandlers } from './events-handler';
+import history from '../../utils/history';
```
dgdavid: Just for the record, probably we could change this in the future if `Buttons` can be `Link` or `NavLink` an the history can be sent in the payload of the action. 

See https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/redux.md

src/state/ducks/room.js
```diff
@@ -27,19 +28,30 @@ const login = (username, room) => {
...
 });
-const logout = () => ({ type: ROOM_LOGOUT });
+
+const logout = () => {
+  return function(dispatch) {
+    janusApi.leaveRoom();
+
+    dispatch({ type: ROOM_LOGOUT });
+
+    history.push('/');
```
dgdavid: As said above, this could change if we are able to manage a more react-router-dom way.

PR #302, opened 2020-02-14
[React] Mirror local camera

ancorgs:
This mirrors the image of the local camera. I just implemented it because I find the lack of such mirroring to be pretty annoying, but I don't know if I followed the right approach taking into account the current architecture and code organization.

Maybe it would make sense to have a `isMirrored` function somewhere and/or to extract the `<video>` tags to some kind of component. But  since I saw similar checks in the modified files, I thought this could be good enough.


dgdavid:
I agree, it was annoying. Thanks for taking care :)

PR #308, opened 2020-03-22
[React] Add thumbnail mode feature

dgdavid:
It allows to enable/disable the video subscription for remote feeds.


src/components/Participant/Participant.js
```diff
@@ -30,15 +34,54 @@ function unsetVideo(videoRef) {
...
+  return (
+    <img
+      data-testid="participant-thumbnail"
+      src={img}
+      onClick={() => dispatchFn(toggleFocus(id, focus))}
+    />
+  );
+}
+
+function Participant({
```
dgdavid: This component should be simplified (e.g., extracting the video logic to its own component, since it's pretty the same logic used for the speaker), but not in the scope of this PR

src/janus-api/action-service.js
```diff
@@ -155,5 +155,13 @@ export const createActionService = (
     feed.setEnabledChannel(type, boolval);
   };
 
+  that.enableVideoSubscriptions = function() {
+    feedsService.remoteFeeds().forEach((f) => f.setVideoSubscription(true));
```
dgdavid: Probably. for the thumbnail mode only not local screen remote feeds should be enabled/disbaled, not all of them. 

What do you think?

src/components/Participant/Participant.js
```diff
@@ -30,15 +34,54 @@ function unsetVideo(videoRef) {
...
+      ref={ref}
+      muted={isPublisher}
+      className={isPublisher && !isLocalScreen ? 'mirrored' : ''}
+      onClick={() => dispatchFn(toggleFocus(id, focus))}
+    />
+  );
+}
+
+function renderImage(id, focus, picture, dispatchFn) {
+  const placeholder = '/placeholder.png';
```
imobachgs: I am not sure whether it will work if the application is deployed to a subdirectory.

src/janus-api/feeds-service.test.js
```diff
@@ -149,6 +149,22 @@ describe('#publisherFeeds', () => {
   });
 });
 
+describe('#remoteFeeds', () => {
+  test('returns an array containing remote feeds', () => {
+    const feedsService = createFeedsService(eventsService);
+    feedsService.add(firstFeed);
+    feedsService.add(secondFeed);
+    expect(feedsService.remoteFeeds()).toEqual(expect.arrayContaining([secondFeed]));
```
imobachgs: `arrayContaining` would succeed if remoteFeeds returns `[firstFeed, secondFeed]` too. See https://jestjs.io/docs/en/expect.html#expectarraycontainingarray.

src/state/ducks/room.js
```diff
@@ -53,29 +54,46 @@ const logout = () => {
   };
 };
 
+const toggleThumbnailMode = () => {
+  return function(dispatch, getState) {
+    let { thumbnailMode } = getState().room;
+
+    thumbnailMode ? janusApi.disableThumbnailMode() : janusApi.enableThumbnailMode();
```
imobachgs: My first thought was to use a `toggleThumbnailMode` function just to be consistent with `toggleVideo` and `toggleAudio`. However, the thumbnail mode is something slightly different (I do not remember if janusApi keeps the current status of the feature). I will check the code.

src/components/Participant/Participant.js
dgdavid: Right. Actually I added it first for testing and I forget to review / comment it.

src/state/ducks/room.js
dgdavid: It was also my first approach, but I ended doing it different because we want to enable or disable de video subscription no matter the current status of them. So, I decided to avoid asking to janusApi. 

We can arrange time to talk about it in a video call / pair programming session if you want.

src/components/Participant/Participant.js
dgdavid: @imobachgs could you please fix this?

imobachgs:
Replaced by #351.

Issue #311, opened 2020-03-24
Write a document describing how to test/devel the react-based version


imobachgs:
There is no documentation about how to test or contribute to the react-based version. We should, at least, write a document explaining how to set it up and use it for testing/development.

PR #312, opened 2020-03-24
Add a README with instructions to run the React based version

imobachgs:
Solves #311.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,68 +1,52 @@
-This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+## Running the React/Redux based branch
 
-## Available Scripts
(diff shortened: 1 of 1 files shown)


pevik:
LGTM, great thanks! The only question I have is whether there is *minimal required* Janus version. Is it the mentioned `v0.8.2`? (latest version is `v0.9.1`).

imobachgs:
> LGTM, great thanks! The only question I have is whether there is _minimal required_ Janus version. Is it the mentioned `v0.8.2`? (latest version is `v0.9.1`).

@pevik Thanks for checking! Well, we recently updated Jangouts to use `v0.8.2`. Actually, the openSUSE packages are based on that version. Anyway, I guess we will need to update it again soon.

PR #326, opened 2020-04-28
[React] WIP: Add action to put the speaker in fullscreen

dgdavid:
## Context

* Few days ago someone asked for such feature in the `#jangouts` IRC channel at the `freenode.net` server
* There is a similar PR open for `master` branch - https://github.com/jangouts/jangouts/pull/185

## Useful links

* https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API
* https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenchange_event

## TODO

- [ ] Unit tests
- [ ] Improve the icon/action placement
- [ ] use the notification system instead an alert?
- [ ] removeEventListener?
     > if so, let's use the useEffect clean up - https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1

## Screenshot

Screenshot_2020-04-28 Jangouts(2)

### DIFF
--- a/src/components/Room/Actions/ToggleFullscreen.js
+++ b/src/components/Room/Actions/ToggleFullscreen.js
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) [2020] SUSE Linux
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE.txt file for details.
+ */
+
+import React, { useEffect, useState } from 'react';
+import { Maximize2 as Maximize, Minimize2 as Minimize } from 'react-feather';
+import { classNames } from '../../../utils/common';
+
+function ToggleFullscreen({elementRef, className}) {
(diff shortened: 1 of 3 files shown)


src/components/Room/Actions/ToggleFullscreen.js
```diff
@@ -0,0 +1,53 @@
...
+
+function ToggleFullscreen({elementRef, className}) {
+  const [usingFullscreen, setUsingFullscreen] = useState(false);
+
+  const Icon = usingFullscreen ? Minimize : Maximize;
+
+  const toggleFullscreen = () => {
+    if (!document.fullscreenElement) {
+      elementRef.current.requestFullscreen()
+        .then(() => setUsingFullscreen(true))
```
dgdavid, line 20: Since there is an event listener for the fullscreenchange, this line is no longer necessary.

Issue #325, opened 2020-04-28
Disconnection of feeds is not properly logged


ancorgs:
When someone quits the room this code is executed, but the feed is always null (likely because it was already destroyed/unregistered by some other piece). https://github.com/jangouts/jangouts/blob/rewrite-using-react-redux/src/janus-api/action-service.js#L61

As a consequence, the event is not logged.

ancorgs:
I was trying to debug this and, indeed, when someone leaves the room two events are generated:

- An `unpublished` one. That's the one that destroys the feed object.
- A `leaving` one, that fails to do its job because the feed is already destroyed.

So I fear this bug affects the master branch as well.

PR #327, opened 2020-04-29
Chat: better filtering of rendering of log messages

ancorgs:
Display the log messages ("John has joined the room", "Mike has muted you", etc.) in its own compact way.

Filter out all the "X has joined the room" messages that are displayed when entering an already populated room (is not them who have joined, but you).

chat-log-messages

As you can see in the screenshots, there are no messages about people leaving the room. That's because of #325

### DIFF
--- a/src/components/Chat/MessagesList/MessagesList.js
+++ b/src/components/Chat/MessagesList/MessagesList.js
@@ -10,9 +10,16 @@ import { useSelector } from 'react-redux';
 import Message from '../Message';
 
 function MessagesList() {
-  const messages = useSelector((state) => state.messages);
   const wrapperRef = useRef(null);
   const messagesRef = useRef(null);
+
+  const filterMessages = (messages) => {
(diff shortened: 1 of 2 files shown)


src/components/Chat/MessagesList/MessagesList.js
```diff
@@ -10,10 +10,17 @@ import { useSelector } from 'react-redux';
...
 function MessagesList() {
-  const messages = useSelector((state) => state.messages);
   const wrapperRef = useRef(null);
   const messagesRef = useRef(null);
 
+  const filterMessages = (messages) => {
+    // Avoid the pointless initial stream of messages like "x has joined the room"
+    // when entering a room in which there are partitipants already. Even if that
+    // means filtering a bit too much in some cases.
+    const first = messages.findIndex((msg) => msg.type !== 'newRemoteFeed');
```
dgdavid, line 20: Just for the record, I think this could be problematic when using the "show/hide chat" feature, but I did not test it yet. I'll do during the week.

PR #333, opened 2020-05-06
[React] Remove pre-commit linters

dgdavid:
I'm who proposed to use husky + prettier + lint-staged (see https://github.com/jangouts/jangouts/pull/237, https://github.com/jangouts/jangouts/pull/244, and https://github.com/jangouts/jangouts/pull/260) to keep the codebase style consistent without too much effort. But now, after giving them a try for this time, I change my mind and I think we should remove them. Main reasons for my current point of view are 

* Sometimes it creates confusion since the committed code could end being quite different from the original written by the developer.
* If I remember correctly, one of them don't work properly under certain circumstances, because the ANSI color escapes codes. In fact, thanks to this problem now I'm aware of the [`NO_COLOR`](https://no-color.org/) initiative.
* You (or at least me) cannot do complex git operations (like modifying a patch on purpose, knowing what are you doing) without risks. Most probably some changes will be lost. Indeed, they [recommend to skip all hooks](https://github.com/typicode/husky#skip-all-hooks-rebase) during a `git rebase`, but... 

So, although I still thinking they are pretty nice tools, maybe we should switch to another (more explicit) approach/workflow. For example, using a [Github Action](https://github.com/features/actions) like https://github.com/samuelmeuli/lint-action

But, first things first. Let's **stop using a pre-commit hook to lint the code**.

Do you agree?

---

BTW, the execution of  `npm install` command after perform those deletions, gave me below output:

```
removed 249 packages and audited 919509 packages in 13.279s
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-src/vendor/janus.js
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,0 @@
-src/vendor/janus.js
--- a/package-lock.json
+++ b/package-lock.json
@@ -2653,15 +2653,6 @@
         "fastq": "^1.6.0"
       }
     },
-    "@samverschueren/stream-to-observable": {
-      "version": "0.3.0",
-      "resolved": "https://registry.npmjs.org/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.0.tgz",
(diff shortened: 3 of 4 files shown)


ancorgs:
> Do you agree?

I do. I'm not against linters in general, but I hate the current setup in which they are executed during the commit phase.

dgdavid:
> in which they are executed during the commit phase.

Lesson learned :roll_eyes: :innocent:

imobachgs:
Thanks! I use [magit](https://magit.vc/) for most of my Git related work and this hook was ruining my workflow :)

dgdavid:
> in which they are executed during the commit phase.

I hope you can forgive me ;)

PR #339, opened 2020-05-29
Final touches for 0.5.0

ancorgs:
This pull request includes the last fixes I wanted to implement before releasing the first lurkers-free version of Jangouts. It's organized into independent commits:

### Provide private_id when subscribing to feeds

Janus Gateway provides a mechanism to correlate subscriber feeds (the mechanism used to consume webRTC streams) to publisher feeds (the mechanism used to join a room). A good Janus citizen should specify the private_if of a valid and active publisher every time it wants to establish a new subscriber feed.

If the room is configured with `require_pvtid`, providing such id becomes mandatory. Combining that setting with `notify_joining` is a good recipe to secure Janus rooms against anonymous lurkers.

Jangouts was already compatible with `notify_joining` (see #337) and now will also honor the private id, making it compatible with `require_pvtid` as well.

### Pass the `withCredentials` argument to Janus

@AdamMajer reported that it was not possible to deploy Jangouts using authentication on the Apache side. Using basic or digest authentication was interfering with the requests to the Janus Gateway server.

Now Jangouts provides the `withCredentials` argument to Janus, which should set the corresponding property of XHR requests. That should make the server authentication work without breaking anything else.

A configuration parameter `janusWithCredentials` was introduced to disable that behavior (it is true by default) in case it turns out to be a problem for some deploy.

### DIFF
--- a/src/app/components/feed/feed-connection.factory.js
+++ b/src/app/components/feed/feed-connection.factory.js
@@ -56,8 +56,11 @@
         pluginHandle.send({"message": register});
       };
 
-      this.listen = function(feedId, pin) {
-        var listen = { "request": "join", "room": roomId, "ptype": "subscriber", "feed": feedId, "pin": pin || "" };
+      this.listen = function(feedId, pin, privateId) {
+        var listen = {
+          "request": "join", "room": roomId, "ptype": "subscriber", "feed": feedId,
+          "pin": pin || "", "private_id": privateId
+        };
         pluginHandle.send({"message": listen});
       };
 
--- a/src/app/components/room/room.service.js
+++ b/src/app/components/room/room.service.js
@@ -81,6 +81,7 @@
         Janus.init({debug: jhConfig.janusDebug});
         that.janus = new Janus({
           server: that.server,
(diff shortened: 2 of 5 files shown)


ancorgs:
@AdamMajer, I would need your confirmation to know whether this fixes your problem.

ancorgs:
Since I suspect and desire 0.5.0 will be the last Angular-based release of Jangouts, I changed my mind about what to include there. There is another thing I would like to implement before marking the release (some visual indication/explanation about how to use the windows resize mode).

So I will remove from this pull request the last commit with the version bump.

ancorgs:
> So I will remove from this pull request the last commit with the version bump.

Done.

dgdavid:
Fine with me

ancorgs:
I'm still missing the feedback from @AdamMajer, but I will merge this anyways to move forward.

AdamMajer:
@ancorgs  Sorry for the late response, but I've now checked and it does indeed work (as expected). Credentials are passed to janus endpoint on Apache side, both BasicAuthentication and Client Certificates seem to work correctly.

PR #346, opened 2020-06-12
Notify speaking while muted

imobachgs:
Adds a notification when the user is speaking but muted. The notification features two buttons:

* "Do not show again" button which allows blocking future "speaking but muted" notifications.
* "Unmute" which allows unmuting.

This notification implements a timeout and it will not pop-up again until 1 minute has passed.

speaking-but-muted

## Notifications infrastructure

The list of blocked notifications has turned into an object which keeps each blocked time and its expiration time. If the expiration is set to `true`, the notification type is blocked indefinitely.

## TODO

I consider the current implementation as good enough and we can refine it in subsequent PRs. There are, at least, two things the I would like to improve:

* Placing of options in the pop-up.
* Resetting the "block" of this notification when the user unmute and mute again (unless the user clicked the 'Do not show again' link).
* Improve the internal representation of the blocklist (using something like a union type would be nice).

### DIFF
--- a/src/components/Notifications/Notifications.test.js
+++ b/src/components/Notifications/Notifications.test.js
@@ -13,7 +13,7 @@ import Notifications from './Notifications';
 const notification1 = createNotification('You have been muted.');
 const notification2 = createNotification('Nobody is listening!');
 const initialState = {
-  notifications: { notifications: [notification1, notification2], blacklist: [] }
+  notifications: { notifications: [notification1, notification2], blocklist: [] }
 };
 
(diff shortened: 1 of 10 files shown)


imobachgs:
> Apart from the location of the actions, I find the label "dismiss" quite confusing. For closing the notification I would expect something like "Ok", "Close", "Accept", "Done" or simply a close button.
> 
> But other than that, LGTM.

OK, let's fix that in another PR.

Thanks!

PR #350, opened 2020-09-25
Do not crash when setStatus contains unknown attributes

imobachgs:
Do not crash when `setStatus` contains unknown attributes. It prevents Jangouts crashing when someone joins to the meeting with an older version.


ancorgs (approved):
LGTM

PR #351, opened 2020-10-03
[React] Thumbnail mode support

imobachgs:
This PR tries to bring back #308. It is still a WIP.

## TODO

- [ ] Integrate the remaining commits from #308 (configurable pictures intervals, bug fixing, etc.).
- [ ] Properly display other participants pictures (`video: true`, although it is ignored).
- [ ] Do not subscribe when a remote participant joins after the thumbnail mode has been enabled.
- [ ] Reproduce video from the speaker.
- [ ] [NTH] Try to reduce the interval to get the first picture.

## Out of Scope

- Split `Participant` into smaller components (display the video, take pictures, display image/picture...).

### DIFF
--- a/src/components/Participant/Participant.js
+++ b/src/components/Participant/Participant.js
@@ -34,6 +34,26 @@ function renderImage(id, picture) {
   }
 }
 
+function takePicture(source, canvas, dispatchFn) {
+  let videoHeight = source.videoHeight;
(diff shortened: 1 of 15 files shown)


imobachgs:
We have decided to postpone this feature. From the UX point of view, we are not sure whether it is a good idea at all.

PR #363, opened 2022-06-28
Update to Janus 1.0 API

imobachgs:
* This PR tries to add support to Janus 1.0 API.
* It should still work with previous versions (0.12.X).

### DIFF
--- a/Dockerfile.janus
+++ b/Dockerfile.janus
@@ -2,12 +2,12 @@ FROM opensuse/leap:15.4
 
 RUN zypper --non-interactive in --no-recommends curl && zypper clean -a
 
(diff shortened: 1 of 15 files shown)


mvidner:
Let me use this PR to bring up my favorite topic: JSDoc :rofl: 
Maybe I am using the wrong tools, but my question is: how do you get an overview of the object that `createRoomService` returns? It looks like an unnamed class. Is that pre-ES6?

imobachgs:
I do not expect JSDoc to work well, as we are not using "modern" JavaScript classes (we are just using factory functions). Perhaps in the future it might change, once the `react-redux` branch is merged into `master`... :-)

ancorgs (approved):
LGTM

PR #365, opened 2022-06-28
Do not display the small video for the main speaker

imobachgs:
When a user is the main speaker, her/his small video should be stopped (replaced by the user icon).

### DIFF
--- a/src/components/Participant/Participant.js
+++ b/src/components/Participant/Participant.js
@@ -39,7 +39,7 @@ function Participant({
   const dispatch = useDispatch();
   const videoRef = React.createRef();
   const cssClassName = `Participant ${focus === 'user' ? 'focus' : ''}`;
-  const showVideo = video || isLocalScreen;
+  const showVideo = (video || isLocalScreen) && !focus;
 
   useEffect(() => setVideo(id, videoRef.current), [id, videoRef, streamReady]);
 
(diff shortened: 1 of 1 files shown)


src/components/Participant/Participant.js
```diff
@@ -39,7 +39,7 @@ function Participant({
   const dispatch = useDispatch();
   const videoRef = React.createRef();
   const cssClassName = `Participant ${focus === 'user' ? 'focus' : ''}`;
-  const showVideo = video || isLocalScreen;
+  const showVideo = (video || isLocalScreen) && !focus;
```
mvidner, line 42: this is simple enough but I'd like to glue some JSDocs for the Participant props onto it :)

src/components/Participant/Participant.js
```diff
@@ -39,7 +39,7 @@ function Participant({
   const dispatch = useDispatch();
   const videoRef = React.createRef();
   const cssClassName = `Participant ${focus === 'user' ? 'focus' : ''}`;
```
mvidner, line 41: ... because here it looks like `focus` is not the simple boolean the one-liner change would suggest

mvidner (changes requested):
Please fill in at least the `focus` property. What does 'user' mean?

```js
/**
 * @param props {object}
 * @param props.id {object} not the superego
 * @param props.username {}
 * @param props.isPublisher {}
 * @param props.isLocalScreen {}
 * @param props.streamReady {}
 * @param props.focus {}
 * @param props.speaking {}
 * @param props.video {}
 */
function Participant({

```

src/components/Participant/Participant.js
imobachgs: Yes, you are right, it is wrong. Let me fix it.

mvidner (approved):

PR #366, opened 2022-06-28
Port improvements from the master branch

ancorgs:
This ports all the improvements of #337 to the react-redux branch. It basically copies those commits one by one... with one exception.

The original #337 also included the changes in the UI. Instead of that, this pull request adds two commits (the last two ones in the list) which expose the new attribute `Feed.connected` in the status of the react app, but without really changing the UI. That's left to a separate pull request.

### DIFF
--- a/src/janus-api/README.md
+++ b/src/janus-api/README.md
@@ -13,8 +13,8 @@ names of its arguments.
 
   * `consentDialog { on }`
   * `createParticipant { id, name, local }`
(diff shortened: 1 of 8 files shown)


imobachgs (approved):

imobachgs (changes requested):
Please, check the use of `isIgnored` in [participants.test.js](https://github.com/jangouts/jangouts/blob/react_lurkers/src/state/ducks/participants.test.js#L23).

imobachgs (approved):

PR #362, opened 2022-06-28
Show the number of the Room's participants

teclator:
Added the number of participants to the Room description

ShowParticipants

PR #367, opened 2022-06-28
Sort rooms by their name

teclator:
It extends #362 to sort the available rooms by their names.

Screenshot from 2022-06-28 16-15-08


ancorgs (approved):
LGTM

PR #368, opened 2022-06-28
Add: npm run jsdoc

mvidner:
Modeled after https://github.com/yast/d-installer/pull/186 but (so far) without the surge.sh publishing


imobachgs (approved):
Thanks a lot!

PR #369, opened 2022-06-29
Add TypeScript support

imobachgs:
Just an experiment to use TypeScript in the `react-redux` branch.

* Adapts `feed.js` (big) and `events-service.js` (simple).
* Some clean-up is still needed.

# To do

- [x] Discuss whether we should use TypeScript
- [x] Clean-up unused methods in `feed.ts`.
- [ ] [Types or interfaces?](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)
- [ ] Improve documentation generation: [better-docs](https://github.com/SoftwareBrothers/better-docs), [jsdoc-plugin-typescript](https://www.npmjs.com/package/jsdoc-plugin-typescript), [typedoc](https://typedoc.org/)...
- [ ] Use TypeScript [getters/setters](https://www.typescriptlang.org/docs/handbook/2/classes.html#getters--setters) (?)


mvidner:
I love types!
So why am I hesitant about this?

I don't know how much work is still left to convert Angular to React, and this adds another conversion from JavaScript to TypeScript. Will that ever get finished to make it to production?

imobachgs:
The good thing about TypeScript is that you do not need to convert the whole codebase. You can adopt it gradually. Sure, you do not have all the benefits, but I see this as a long-term goal.

ancorgs (approved):
I know nothing about TypeScript, but I trust this looks ok.

Issue #331, opened 2020-04-30
[bug] [branch: react-redux] no new chat notifications


Wurzelmann:
Regardless if the chat is hidden or shown, there are no notifications about new messages (like the number of new messages next to the chat icon or small indicator like a coloured circle or something similar), which were part of the "classic" jangouts interface, if I'm not mistaken.

joseivanlopez:
* In some cases, emojis are not correctly rendered.
* Emoji peeker is missing.

joseivanlopez:
Some chat improvements were done in https://github.com/jangouts/jangouts/pull/383.

ancorgs:
Now we have the counter for new (unread) chat messages. Thus, I'm closing this issue. @joseivanlopez feel free to open a separate one for the other things you mentioned in your comment, they are not related to the original scope of the issue.

PR #374, opened 2022-06-29
Recover user settings

joseivanlopez:
## Problem

Last used username and room are lost from the login form after refreshing.

* https://github.com/jangouts/jangouts/issues/359

## Solution

Introduce `UserSettings` class, which stores and retrieves its data from browser local storage.

PR #383, opened 2022-06-30
Improve chat

joseivanlopez:
Some chat improvements:

* Layout is remembered (open/closed chat).
* Show number of unread chat messages.

This PR fixes part of https://github.com/jangouts/jangouts/issues/331.


src/components/Chat/Toggler/Toggler.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function Toggler() {
+  const dispatch = useDispatch();
+  const { settings } = useSelector((state) => state.room);
+
+  let title;
+  let classes;
+
+  if (settings.chatOpen) {
```
mvidner, line 28: I don't understand how `settings` can have a value now.
It comes from `useSelector`, which is a library function, not having access to any of our data, and we pass it only a trivial lambda.
Redux magic?

mvidner: Well maybe it's why the tests fail too?

imobachgs:
Actually, that's a good question. And yes, there is some Redux magic there. I will try to explain what I know. @joseivanlopez might be interested in this topic too, as we were talking about that recently.

First of all, I would like to recommend installing the [Redux DevTools](https://github.com/reduxjs/redux-devtools). IMHO, it is a must-have when working with Redux, as it allows you to see Redux actions and how they change the state.

Now, let's start from the beginning. As you know, Redux keeps our application state in a single object. Initially, it looks like this:

```javascript
{
  room: {
    settings: {},
    loggedIn: false,
    loggingIn: false
  },
  participants: {},
  messages: [],
  config: {},
  notifications: {
    notifications: [],
    blocklist: {}
  }
}
```

You can read the [Initializing State](https://redux.js.org/usage/structuring-reducers/initializing-state) to know more about how the store is initialized. In our case, we pass the initial state of each section to its corresponding reducer. See [src/state/ducks/room.js](https://github.com/jangouts/jangouts/blob/1e5ede00368f321db9128470d99ae39463431ab1/src/state/ducks/room.js#L82) for an example. When the application starts, Redux fires an `@@INIT` action which causes all reducers to run and initialize the state.

Later, the state might look like this:

```javascript
{
  room: {
    settings: {
      _username: 'imo-laptop',
      _roomId: '2003'
    },
    loggedIn: false,
    loggingIn: false
  },
  participants: {},
  messages: [],
  config: {
    janusWithCredentials: true,
    joinUnmutedLimit: 3,
    videoThumbnails: true,
    janusServer: 'wss://li2023-182.members.linode.com/janus/'
  },
  notifications: {
    notifications: [],
    blocklist: {}
  }
}
```

According to [useSelector documentation](https://react-redux.js.org/api/hooks#useselector), it *allows you to extract data from the Redux store state, using a selector function*. So it receives the whole state and it just returns the part you are interested in.

Finally, you might wonder why `_username` and `_roomId` are prepended with an underscore. The reason is that they are private attributes of the `UserSettings` class. And that takes me to the next point: *it is highly recommended that you only put plain serializable objects, arrays, and primitives into your store* ([source](https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state)). I am sorry, @joseivanlopez, but I forgot about that detail 😞. But we can adapt #374 easily 😉.

src/components/Chat/Toggler/Toggler.js
imobachgs: Actually, that's a good question. And yes, there is some Redux magic there. I will try to explain what I know. @joseivanlopez might be interested in this topic too, as we were talking about that recently.

First of all, I would like to recommend installing the [Redux DevTools](https://github.com/reduxjs/redux-devtools). IMHO, it is a must-have when working with Redux, as it allows you to see Redux actions and how they change the state.

Now, let's start from the beginning. As you know, Redux keeps our application state in a single object. Initially, it looks like this:

```javascript
{
  room: {
    settings: {},
    loggedIn: false,
    loggingIn: false
  },
  participants: {},
  messages: [],
  config: {},
  notifications: {
    notifications: [],
    blocklist: {}
  }
}
```

You can read the [Initializing State](https://redux.js.org/usage/structuring-reducers/initializing-state) to know more about how the store is initialized. In our case, we pass the initial state of each section to its corresponding reducer. See [src/state/ducks/room.js](https://github.com/jangouts/jangouts/blob/1e5ede00368f321db9128470d99ae39463431ab1/src/state/ducks/room.js#L82) for an example. When the application starts, Redux fires an `@@INIT` action which causes all reducers to run and initialize the state.

Later, the state might look like this:

```javascript
{
  room: {
    settings: {
      _username: 'imo-laptop',
      _roomId: '2003'
    },
    loggedIn: false,
    loggingIn: false
  },
  participants: {},
  messages: [],
  config: {
    janusWithCredentials: true,
    joinUnmutedLimit: 3,
    videoThumbnails: true,
    janusServer: 'wss://li2023-182.members.linode.com/janus/'
  },
  notifications: {
    notifications: [],
    blocklist: {}
  }
}
```

According to [useSelector documentation](https://react-redux.js.org/api/hooks#useselector), it *allows you to extract data from the Redux store state, using a selector function*. So it receives the whole state and it just returns the part you are interested in.

Finally, you might wonder why `_username` and `_roomId` are prepended with an underscore. The reason is that they are private attributes of the `UserSettings` class. And that takes me to the next point: *it is highly recommended that you only put plain serializable objects, arrays, and primitives into your store* ([source](https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state)). I am sorry, @joseivanlopez, but I forgot about that detail 😞. But we can adapt #374 easily 😉.

joseivanlopez: Fixed.

dgdavid (approved):
LGTM

I left two neat pickings, but you can perfectly ignore them :stuck_out_tongue_closed_eyes:

PR #384, opened 2022-06-30
Fix Feed constructor

imobachgs:
Fix a problem introduced when working on #369 which caused some Feed properties to not be properly initialized. It caused incoming participants to not be shown.

### DIFF
--- a/src/janus-api/models/feed.ts
+++ b/src/janus-api/models/feed.ts
@@ -30,8 +30,8 @@ const statusAttrs = ['name', 'speaking', 'audio', 'video', 'picture'];
  * Feed (?)
  */
 export class Feed {
-  id: number = 0;
-  display: string;
+  id: number;
+  display: string | null;
   publisher: boolean;
   localScreen: boolean;
   ignored: boolean;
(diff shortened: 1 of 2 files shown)


src/janus-api/models/feed.ts
```diff
@@ -48,12 +48,12 @@ export class Feed {
...
-    this.publisher = isPublisher;
-    this.localScreen = isLocalScreen;
-    this.ignored = ignored; // is still used?
-    this.connection = connection;
+    this.id = id || 0;
+    this.display = display || null;
+    this.publisher = isPublisher || false;
+    this.localScreen = isLocalScreen || false;
+    this.ignored = ignored || false; // is still used?
+    this.connection = connection || null;
```
dgdavid, line 56: NP: what about using the [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) here?

dgdavid (approved):
Looks good. Thanks for fixing it!

PR #386, opened 2022-07-01
Fix configuration loading

### LINKED ISSUE #385: Cannot load the configuration file
Jangouts supports a few configuration options using a `config.json` file which is retrieved at runtime. This mechanism does not work in the `react-redux` branch.

imobachgs:
Fixes #385. By the way, I took the opportunity to simplify the code by using the [Fetch API](https://developer.mozilla.org/es/docs/Web/API/Fetch_API).

### DIFF
--- a/src/utils/config.js
+++ b/src/utils/config.js
@@ -50,30 +50,22 @@ const defaultJanusServer = () => {
 const replacePlaceholders = (text) => text.replace('%{hostname}', window.location.hostname);
 
 /**
- * Parses a piece of JSON returning an empty object if it is not valid.
- *
- * @param {string} text Configuration (in JSON format).
- * @returns {Object}
- */
-const parseJSON = (text) => {
-  try {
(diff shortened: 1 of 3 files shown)


src/utils/config.js
```diff
@@ -50,63 +50,43 @@ const defaultJanusServer = () => {
...
 
-const READY_STATE_DONE = 4;
-const STATUS_SUCCESS = 200;
-
 /**
- * Fetches and parses the configuration from the `config.json` file.
+ * Retrieves the configuration
  *
- * @returns {Object}
+ * It tries to get a custom configuration from `/config.json`. If a valid configuration
```
dgdavid, line 79: For other readers: example available in https://github.com/jangouts/jangouts/blob/215b1b107d332ea0108869213755dd8573cd7c98/public/config.json.sample

dgdavid (approved):
LGTM :+1:

PR #388, opened 2022-07-01
Avoid participant flickering

imobachgs:
## Problem

Each time a participant start to speak their video flickers because the video got detached and re-attached due to a component re-render.

## (Temporary) Solution

From the time being, we have extracted the logic in charge to set video stream to a `ParticipantVideo` component, which is wrapped in a `React.memo`. It helps and the flickering is away, but maybe we can find a better solution in next iterations.


ancorgs (approved):
LGTM

PR #391, opened 2022-07-01
Improve unread messages counter

dgdavid:
It has room for improvements yet, but at least it is a bit more readable now.

Screenshot 2022-07-01 at 17-12-19 Jangouts


ancorgs (approved):
Much better now!

PR #395, opened 2023-02-01
Stop tracks when disconnecting (fixes ##394)

### LINKED ISSUE #394: When quiting a room, the camera remains activated
If you use the icon "Leave the room", you end up in the login form... but with the camera still on.

ancorgs:
When leaving the room, stop all the tracks so the browser is aware of the disconnection and turns off the physical indicators (like the camera light).

Fixes #394

### DIFF
--- a/src/janus-api/models/feed.ts
+++ b/src/janus-api/models/feed.ts
@@ -224,6 +224,19 @@ export class Feed {
       this.speakObserver.destroy();
     }
     this.connection = null;
+    this.removeStream();
+  };
+
+  /**
+   * Unsets the stream from the feed, stopping all its tracks.
+   */
+  private removeStream() {
+    if (this.stream) {
+      let tracks = this.stream.getTracks();
+      for(let i in tracks) {
+        tracks[i].stop();
+      }
+    }
     this.stream = null;
   };
 


ancorgs:
Not adding a unit test because the only one I can come with would be all mocking and basically reproducing the implementation.

imobachgs (approved):

src/janus-api/models/feed.ts
```diff
@@ -224,6 +224,19 @@ export class Feed {
...
+    this.removeStream();
+  };
+
+  /**
+   * Unsets the stream from the feed, stopping all its tracks.
+   */
+  private removeStream() {
+    if (this.stream) {
+      let tracks = this.stream.getTracks();
+      for(let i in tracks) {
```
joseivanlopez, line 236: Maybe this also works:

~~~
tracks.map(t => t.stop());
~~~

imobachgs: I would use `forEach` because you are not "mapping" at all.

joseivanlopez: ah, right. I  only wanted to point that we can "functionally" iterate.

PR #397, opened 2023-02-02
Calculate participant sizes with Javascript

ancorgs:
In the react-redux branch we tried to calculate the size of the participants only with CSS. But the behavior is not the desired one. We wanted it to behave similar to the Angular branch, which seems impossible using pure CSS.

This pull request brings back the logic used in the Angular branch, but with a twist. Instead of setting the width and height of each participant directly with Javascript, this uses Javascript to calculate the desired width of the participants and stores the value a CSS variable. By default, the algorithm assumes a square representation of each participant, but it can be configured (using the `--partHeightFactor` CSS variable) to consider any other proportion (eg. 4:3). The algorithm can be further configured with other CSS variables like `---partMaxSlotWidth`, `--partMinSlotWidth` and `--partStepSlotWidth` to define the range of sizes and like `--partHeightExtra` and `--partWidthExtra` to fine-tune the calculation when `--partHeightFactor` is different from the default 1. 

This PR also includes some changes in the CSS that rely on the mentioned mechanism to provide a list of participants with a fixed 4:3 ratio in the videos, similar to the one in the Angular branch.

hablando

### DIFF
--- a/src/components/Participant/Participant.js
+++ b/src/components/Participant/Participant.js
@@ -39,7 +39,8 @@ const Video = ({id, focus, streamReady, showVideo, isPublisher, isLocalScreen}) 
       autoPlay
       className={classNames(
         showVideo || 'hidden',
-        isPublisher && !isLocalScreen && 'mirrored'
+        isPublisher && !isLocalScreen && 'mirrored',
+        "max-h-full"
       )}
       onClick={() => dispatch(toggleFocus(id, focus))}
     />
@@ -74,14 +75,16 @@ function Participant({
   return (
     <div
       className={classNames(
-        'participant relative group p-1 border-2 bg-white',
+        'participant relative group bg-white',
         'transition duration-150 ease-in-out',
         focus || 'border-white',
         focus && 'border-secondary shadow-md',
         speaking && 'border-green-300'
       )}
-      style={{width: "calc(var(--partSlotWidth) - 6px", height: "calc(var(--partSlotWidth) - 6px)", margin: "3px"}}>
-      <div className="relative bg-gray-100">
+      style={{width: "calc(var(--partSlotWidth) - 6px)", margin: "3px", padding: "4px", "border-width": "2px"}}>
+      <div
+        className="relative bg-gray-100"
+        style={{"height": "calc((var(--partSlotWidth) - 18px) * 0.75)"}}>
         <ParticipantVideo 
           id={id}
           focus={focus}
@@ -94,7 +97,7 @@ function Participant({
       </div>
       <div className="flex items-center p-1 bg-gray-200">
         <ParticipantActions participantId={id} />
-        <div className="text-sx md:text-sm whitespace-no-wrap truncate">{username}</div>
+        <div className="text-sm whitespace-no-wrap truncate">{username}</div>
       </div>
     </div>
   );
--- a/src/components/Participant/Participant.js
+++ b/src/components/Participant/Participant.js
@@ -74,12 +74,13 @@ function Participant({
   return (
     <div
       className={classNames(
-        'relative group p-1 border-2 bg-white',
+        'participant relative group p-1 border-2 bg-white',
         'transition duration-150 ease-in-out',
         focus || 'border-white',
         focus && 'border-secondary shadow-md',
         speaking && 'border-green-300'
-      )}>
+      )}
+      style={{width: "calc(var(--partSlotWidth) - 6px", height: "calc(var(--partSlotWidth) - 6px)", margin: "3px"}}>
       <div className="relative bg-gray-100">
         <ParticipantVideo 
           id={id}
--- a/src/components/Participants/Participants.js
+++ b/src/components/Participants/Participants.js
@@ -14,13 +14,8 @@ import Participant from '../Participant';
  * See setParticipantVars
  */
 function getCssConfig(element, property, def) {
-  let value = element.style.getPropertyValue(property);
-
-  if (value === "") {
-    return def;
-  } else {
-    return Number(value);
-  }
+  const value = element.style.getPropertyValue(property);
+  return value === "" ? def : Number(value);
 }
 
 /*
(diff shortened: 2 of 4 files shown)


src/components/Participant/Participant.js
```diff
@@ -74,12 +74,13 @@ function Participant({
...
       className={classNames(
-        'relative group p-1 border-2 bg-white',
+        'participant relative group p-1 border-2 bg-white',
         'transition duration-150 ease-in-out',
         focus || 'border-white',
         focus && 'border-secondary shadow-md',
         speaking && 'border-green-300'
-      )}>
+      )}
+      style={{width: "calc(var(--slotSize) - 6px", height: "calc(var(--slotSize) - 6px", margin: "3px"}}>
```
dgdavid, line 83: Missing parenthesis for closing the `calc`s. Is it working?

src/components/Participant/Participant.js
```diff
@@ -74,12 +74,13 @@ function Participant({
...
       className={classNames(
-        'relative group p-1 border-2 bg-white',
+        'participant relative group p-1 border-2 bg-white',
         'transition duration-150 ease-in-out',
         focus || 'border-white',
         focus && 'border-secondary shadow-md',
         speaking && 'border-green-300'
-      )}>
+      )}
+      style={{width: "calc(var(--partSlotWidth) - 6px", height: "calc(var(--partSlotWidth) - 6px)", margin: "3px"}}>
```
dgdavid, line 83: Still missing a bracket for closing `width`'s `calc`.

src/components/Participant/Participant.js
```diff
@@ -74,13 +75,16 @@ function Participant({
...
-        'relative group p-1 border-2 bg-white',
+        'participant relative group bg-white',
         'transition duration-150 ease-in-out',
         focus || 'border-white',
         focus && 'border-secondary shadow-md',
         speaking && 'border-green-300'
-      )}>
-      <div className="relative bg-gray-100">
+      )}
+      style={{width: "calc(var(--partSlotWidth) - 6px)", margin: "3px", padding: "4px", "border-width": "2px"}}>
```
joseivanlopez, line 84: Question: Could we place these width and height attributes with `calc` directly in the css file?

dgdavid: Yes, we can place them in a `.participant` CSS class which we can define in the _components layer_ at https://github.com/jangouts/jangouts/blob/5ab3baa06a64295461f14c5ed8bc8a1dd018796a/src/styles/index.css#L24-L48

 I guess @ancorgs didn't go for it because we are (almost) not defining CSS classes by ourselves and using Tailwind instead. BTW, I'd like to change that in the future. Wanna you join?

dgdavid: > we can place them in a `.participant` CSS class 

If so, please add documentation there too. At least a reference to the `.js` file where its meaning and the reason for its existence is explained.

src/components/Participants/Participants.js
```diff
@@ -5,21 +5,101 @@
...
+ */
+function getCssConfig(element, property, def) {
+  let value = element.style.getPropertyValue(property);
+
+  if (value === "") {
+    return def;
+  } else {
+    return Number(value);
+  }
+}
```
dgdavid, line 24: NP: I'd use a ternary operator over here. A matter of taste.

```js
return value === "" ? def : Number(value)
```

src/components/Participants/Participants.js
```diff
@@ -5,21 +5,101 @@
...
+  let totalHeight = div.offsetHeight;
+  let qty = div.querySelectorAll(".participant").length;
+
+  let max = getCssConfig(div, "--partMaxSlotWidth", 160);
+  let min = getCssConfig(div, "--partMinSlotWidth", 80);
+  let step = getCssConfig(div, "--partStepSlotWidth", 4);
+
+  let heightFactor = getCssConfig(div, "--partHeightFactor", 1);
+  let heightExtra = getCssConfig(div, "--partHeightExtra", 0);
+  let widthExtra = getCssConfig(div, "--partWidthExtra", 0);
```
dgdavid, line 51: NP: it's a common practice to use `const` instead of `let` when you're not going to reassign the variable.

src/components/Participants/Participants.js
```diff
@@ -5,21 +5,101 @@
...
+  useEffect(() => {
+    setParticipantVars(mainRef.current);
+  }, [orderedParticipants.length]);
+
   return (
-    <div className="grid items-center grid-cols-3 lg:grid-cols-5 gap-2 row-gap-3 p-2">
+    <div
+      ref={mainRef}
+      className="h-full w-full flex flex-row flex-wrap content-start overflow-x-hidden overflow-y-auto"
+      style={{"--partWidthExtra": 18, "--partHeightExtra": 46, "--partHeightFactor": 0.75}}>
```
dgdavid, line 102: If you are going to follow the @joseivanlopez's suggestion and move the Participant style to a CSS class, please add a `.participants` CSS class for this div and move its style there too.

dgdavid (commented):
So far LGTM. Just minor comments.

src/components/Participants/Participants.js
ancorgs: Not for now. I leave that as an exercise for CSS lovers.

src/components/Participants/Participants.js
ancorgs: Done.

src/components/Participants/Participants.js
ancorgs: Done (I hope).

src/setupTests.js
```diff
@@ -20,6 +20,13 @@ import reducers from './state/ducks';
 // in jsdom yet. Related to https://github.com/jsdom/jsdom/issues/1695
 Element.prototype.scrollTo = jest.fn();
 
+// Mock the browser's observer used to resize the participant thumbnails
+class MockObserver {
+  observe() { jest.fn() }
+  disconnect() { jest.fn() }
+}
+window.ResizeObserver = MockObserver;
```
dgdavid, line 28: NP: Honestly, I'm not sure but maybe would be better to wrap this in a `beforeAll` and then do a [`.mockRestore`](https://jestjs.io/docs/mock-function-api#mockfnmockrestore) in a `afterAll`. Just in case.

Anyway,  PR LGTM as it is, so you can go ahead an merge.

dgdavid (approved):
LGTM, just a minor comment (for the future).

PR #399, opened 2023-02-03
Redistribute the layout in big screens

ancorgs:
I wide screens, giving half of the horizontal space to the chat looks like too much. This reduces that to only one third if the browser width is 1024px or bigger.

The first row in the table below shows the result in a browser windows => 1024px.

The second row is 1024 > browser > 640.

The third one is browser < 640px.

| With chat             |  Hidden chat |
:-------------------------:|:-------------------------:
|
|
|


dgdavid (approved):
Ok, let's go.

Issue #392, opened 2022-07-04
Offer a button to (un)mute and enable/disable the video in the toolbar


imobachgs:
If you want to unmute yourself, for instance, to say "goodbye" at the end of a call, it is hard to click on the "unmute" button because it is moving as the people are leaving the room. I would be good to have a button to (un)mute and enable/disable the video at the top bar.

ancorgs:
Being worked at https://github.com/jangouts/jangouts/pull/400

imobachgs:
Closed by #400.

PR #400, opened 2023-02-03
React UI tweaks

ancorgs:
Two UI improvements:

- Manage the focus by clicking on the whole preview area of the participant in the roster, not necessarily in the video element. Removing the focus becomes possible again.
- Add buttons to toggle video and audio in the header. 

Screenshot from 2023-02-03 16-08-42

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -19,7 +19,7 @@ function Header({ children }) {
       <div className="flex w-1/5 py-1 justify-center">
         <Logo className="h-6 w-auto" />
       </div>
-      <div className="flex w-2/5 justify-end">
+      <div className="flex gap-1 w-2/5 justify-end">
         {/* General actions */}
         {children}
         <RoomActions className="ml-4 p-px" />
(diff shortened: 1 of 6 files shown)


ancorgs:
This should fix #392

ancorgs:
It looks much better with the new commits from @dgdavid. But I think the toggle chat look&feel could be more consistent with the mute/video ones.

ancorgs:
LGTM, in any case, after the changes.

dgdavid:
> It looks much better with the new commits from @dgdavid. But I think the toggle chat look&feel could be more consistent with the mute/video ones.

Consistency is a bit tricky here. At least from my point of view.

The reason is that I see at them as different group of toggle actions. For the mic and camera, we want to visually communicate that they are on or off, using the green color for the icons when active and the red color for the background when inactive. Hopefully, this can help the user to quickly realize the state of those inputs. In the case of the chat… we don't matter if you wanted to read it or not. It just uses the white background when active…

But sure, in next rounds we can think about it and bring a better solution in the shake of consistency.

dgdavid (approved):

PR #402, opened 2023-02-03
Another couple of small UI tweaks

ancorgs:
Review by commit. They should be self-explanatory.


dgdavid:
It looks good, 

Screenshot from 2023-02-03 16-14-06


Just wondering if worth adding `@ YaST #1 room` or similar for better understanding of the label meaning

ancorgs:
> Just wondering if worth adding `@ YaST #1 room` or similar for better understanding of the label meaning

We don't have it in the Angular branch and I have never felt that's a problem. You should recognize at first sight the name of the room you are in. Isn't it?

PR #403, opened 2023-02-03
Fix problem with icon color of toggle actions

dgdavid:


ancorgs (approved):
I verified this fixes the problem in production.

PR #416, opened 2023-02-13
Better links in chat

### LINKED ISSUE #412: Not all URLs are turned into links in the chat
We are using Interweave that, by default, only turn into links URLS [from a small set of TLDs](https://github.com/milesj/interweave/blob/master/website/docs/exts/autolink.mdx#tld-support). It feels like too much. We need to either:

- Expand that list of TLDs
- Disable TLD validation

ancorgs:
This should fix #412

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -62,6 +62,10 @@
     border-top-width: 1px;
   }
 
+  .chat-wrapper a {
+    text-decoration: underline;
+  }
+
   .chat-position-toggler {
     --icon-size: 24px;
     --half-icon-size: calc(var(--icon-size) / 2);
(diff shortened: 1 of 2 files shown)


src/styles/index.css
```diff
@@ -62,6 +62,10 @@
     border-top-width: 1px;
   }
 
+  .chat-wrapper a {
+    text-decoration: underline;
```
dgdavid, line 66: Of course, it will work. But I was thinking more in using the [`revert`](https://developer.mozilla.org/en-US/docs/Web/CSS/revert) CSS keyword value in order to use the default specified by the browser in its UA stylesheet by now and bypassing the Tailwind reset.

```css
.chat-wrapper a {
  color: revert;
  text-decoration: revert;
}
```

This makes the link not only to be underlined, but also use the browser defined colors for (if I'm not wrong) a link, a link:hover, a link:visited, etc.

But as you wish, we can go ahead just underlined them by now.

dgdavid (approved):

PR #405, opened 2023-02-08
Some UI adjustments

dgdavid:
Just a few layout improvements based on what I saw in a group meeting today.

Namely, 

* [Improve participants alignment by distributing space evenly](https://github.com/jangouts/jangouts/commit/d0a31e2c227c0ab5d17dd95852fd7d7a8433c0b1)
  
  <em>See the amount of available space at the right of participants list</em>

  | Before | After |
  |:-:|:-:|
  | Screenshot from 2023-02-08 12-12-30 | Screenshot from 2023-02-08 12-16-24 |

* [Better alignment of participant video](https://github.com/jangouts/jangouts/commit/2cd894738e7b026134596d30b9b588e3283b3924)

  <em>I tested that it works in both sides, x and y axis.</em>

  | Before | After |
  |:-:|:-:|
  | Screenshot from 2023-02-08 12-24-25 | Screenshot from 2023-02-08 12-25-50 |

* [Keep improving participants alignment](https://github.com/jangouts/jangouts/commit/6a16000778e4c7a79dee832b10440bb15ed6420e)

  <em>By changing from Flexbox to CSS Grid to avoid "weird" (for us) behavior when distributing space evenly in the last row. To know more, see https://travishorn.com/some-ways-to-align-the-last-row-in-a-flexbox-grid-720f365dcb16 for example.</em>

  | Before | After |
  |:-:|:-:|
  | Screenshot from 2023-02-08 12-41-29 | Screenshot from 2023-02-08 12-41-20 |

PR #417, opened 2023-02-14
Fix participant background color

dgdavid:
Change made at commit https://github.com/jangouts/jangouts/pull/405/commits/2cd894738e7b026134596d30b9b588e3283b3924 as part of #405 makes the participant icon looks very ugly. So, this change is about using the dark background only when the participant is rendering a video.

|Before | After |
|-|-|
|Screenshot 2023-02-14 at 10-20-00 Jangouts | Screenshot 2023-02-14 at 10-19-37 Jangouts |


ancorgs (approved):
LGTM

PR #419, opened 2023-02-15
Add backwards compatible URL to rooms

ancorgs:
This addresses #414.

This approach of just having a `<Route>` for each type of URL seems to work just fine in my experiments. From the user point if view it looks just like a redirection.

In fact, I tested several approaches with `<Navigate>` and other redirection techniques but they were more complex and I didn't appreciate any advantage.


dgdavid (approved):
LGTM

PR #421, opened 2023-02-17
Polish configuration

ancorgs:
## Problem

The configuration in the `react-redux` branch was confusing for several reasons.

1) Both the `janusAPI` and `fetchConfig` contained logic to calculate the default values of the settings, especially the URL of the Janus server. The logic was kind of equivalent but not identical.

2) For historical reasons, there were two settings called `janusServer` and `janusServerSSL`. That's confusing and unnecessary. Especially taking into account that SSL is a must anyways, since modern browsers will not open WebRTC connections unless SSL is used.

3) The setting `janusDebug` got missing in the transition from the  `master` (Angular) branch to the `react-redux` one.

4) Moreover, there was a bug. `fetchConfig` should replace the placeholder `%{hostname}` in the config file with its corresponding value. But only the first occurrence was replaced.

## Solution

Now `janusAPI` does not calculate any default value. That responsibility is delegated to the application (`fetchConfig` in our case) and `janusAPI` requires the final configuration to be given.

Adapted the logic in `fetchConfig` to have the best parts of the two previous calculations.

Removed `janusServerSSL`. Now there is only one parameter to set the URL(s) of the Janus server in all cases. Note it can be an array with several URLs to be all tried.

Bring back a functional `janusDebug` setting.

Fixed the bug in parsing the config file. Now all the occurrences of the placeholder are replaced.

### DIFF
--- a/public/config.json.sample
+++ b/public/config.json.sample
@@ -1,5 +1,6 @@
 {
   "janusServer": null,
+  "janusDebug": false,
   "janusWithCredentials": true,
   "joinUnmutedLimit": 3,
   "videoThumbnails": true
--- a/src/janus-api/index.js
+++ b/src/janus-api/index.js
@@ -26,7 +26,7 @@ export default (function() {
   };
 
   that.setup = function(
-    { janusServer, janusServerSSL, janusWithCredentials, joinUnmutedLimit, videoThumbnails },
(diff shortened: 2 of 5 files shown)


dgdavid (approved):
LGTM

PR #422, opened 2023-02-19
Avoid React warning about css property

ancorgs:
This brings not visible change in the UI, but silents a warning in the javascript console.


dgdavid (approved):
Thanks!

Issue #413, opened 2023-02-11
Wrong (or missing) PIN is poorly handled


ancorgs:
If a room requires PIN but is not correctly entered, no error is reported. Instead, the room layout loads but the user is not allowed to connect to Janus... which is far from obvious.

ancorgs:
When the connection fails due to a wrong or missing PIN, janusAPI emits a `reportError` event. But looks like the UI does nothing about it.

ancorgs:
And this is how we handle it in the Angular branch. Nothing too impressive but quite functional.

https://github.com/jangouts/jangouts/blob/33ac5ff05ae77b4bf1b2114431d6558041deb5c8/src/app/room/room.controller.js#L39

ancorgs:
Fixed (or at least greatly mitigated) by #425

PR #425, opened 2023-02-20
Handle wrong PIN

ancorgs:
Mitigates #413 

Currently deployed at https://li2023-182.members.linode.com/beta/

The React part was somehow prepared to handle a login error (see #427 for more details about that "somehow"), but the `janusApi` always resolved the `enterRoom` promise, no matter whether the operation succeeded or not. With this patch, it rejects the promise in case of error.

In addition, this includes fixes to the React part to cleanup previous errors when a new login succeeds and to memoize the username and the room id if the user skips the login form.

The pull request also includes a fix for a race condition that caused some queries to `janusAPI` to fail. It became especially visible when entering directly the URL of a room with pin (it now redirects back to the login form and the query to populate the list of rooms failed there due to the race condition).

This is probably all the best we can do without tackling #427 (apart from the redirection flickering caused by that, it works as expected in all situations).

### DIFF
--- a/src/components/Room/Room.js
+++ b/src/components/Room/Room.js
@@ -28,6 +28,9 @@ function Room() {
     const searchParams = new URLSearchParams(location.search);
     const username = searchParams.get('user') || randomUsername();
     // TODO: get username from local storage
+
+    // Save settings, in case we don't come from the login form
+    dispatch(roomActions.saveSettings({...room.settings, username, roomId}));
     dispatch(roomActions.login(username, roomId));
   }, []);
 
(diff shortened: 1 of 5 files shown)


dgdavid (approved):

ancorgs:
I updated this pull request and also the testing instance deployed at https://li2023-182.members.linode.com/beta/

Please re-review.

dgdavid (approved):
Still looking good. Actually better. I have done some tests in the beta instance and it works 💯

PR #429, opened 2023-03-01
Prepare the react-redux branch to become the new master

ancorgs:
We are ready to declare the `react-redux` version as the official Jangouts 0.6.0.

This pull request gives the final touches to the branch. After merging this we could simply merge `react-redux` into master and do the release.

First of all, this brings back all the documentation that is available in master but was missing in the `react-redux` branch. The information (with the exception of the screenshot, that is reused as-is) has been adapted and proof tested. I followed all the steps at README.md and at DEVELOPMENT.md in a clean machine that didn't have Janus or npm initially installed. Everything worked by just following the instructions (using our Dockerfile for Janus).

This also increases the version at the `package.json` and `package-lock.json` files. I just modified the version number there in both files by hand. I don't know whether there is a cleaner approach.


imobachgs (approved):
Congrats!

PR #437, opened 2023-11-10
Stop using tailwindcss

dgdavid:
I wanted to start working on layouts during HackWeek 23. The plan was to rely on https://github.com/bvaughn/react-resizable-panels to do so. 

It was also my wish to add a new theme, but I felt getting rid of tailwindcss was the first step for both cases. While I suggested using it when migrating from AngularJS to React to _rapidly build_ the UI, I've changed my mind now that I'm back in the web development world. I think it is not an appropriate tool for this project, although this is a personal vision of someone that prefers [declarative design over imperative one](https://adactio.com/journal/18982).

I know I should elaborate the rationale a bit more. Unfortunately, I am a bit drained when it comes to writing long PR descriptions. As such, I will try to do it in the future, but I make absolutely no promises. In short, I see components tightly coupled to UI definitions, which can be an obstacle for building just one style agnostic markup and letting the _browser CSS engine_ to do the work based in a few CSS rules. Think about CSS Grid with named areas, for example.

---

About the result of this PR: the look&feel is _almost_ the same as before, but without using tailwindcss. Changes made the current layout less usable in the small devices, though. Other small things had improved, like notifications. In any case, there is still work to do in subsequent PRs, reason why the PR is open against `next` branch. Do not be scared by the (maybe temporary) CSS rules. They are going to be improved before opening a PR against `master`. In fact, the _src/styles/migration.css_ file should disappear.


dgdavid:
Due to the limited chances (if any) of me investing more time in this beloved project, I am closing this PR. Merging it without a continuity commitment does not make sense.

I enjoyed contributing anyway. It was a great experience. Thank you so much!

